PR #24, opened 2020-04-25
Adds ingest.FromReader(), some doc cleanup.

element-of-surprise:
Adds a new ingest.FromReader() method so that it is easier to write directly from memory.    While this can be done with Stream(), stream is still considered non-prod and must be enabled on a cluster.  This method can be used by any cluster.

Also added some doc fixes.


README.md
```diff
@@ -112,6 +112,23 @@ some prerequisite knowledge of acceptable data formats, mapping references, ...
 
 That documentation can be found [here](https://docs.microsoft.com/en-us/azure/kusto/management/data-ingestion/)
 
+Kusto's ingestion service makes no guarantees on when the data will show up in the table and is optimized for
+large chunks of data and not small uploads at a high rate.
+
+If ingesting data from memory, it is suggested that you stream the data in via FromReader() passing in the reader
+from an io.Pipe().  The data will not be begin ingestion until the writer closes.
```
danield137, line 119:
```suggestion
from an io.Pipe(). The data will not begin ingestion until the writer closes.

```

kusto/ingest/internal/filesystem/filesystem.go
```diff
@@ -91,6 +105,53 @@ func (i *Ingestion) Local(ctx context.Context, from string, props properties.All
...
+	if len(resources.Queues) == 0 {
+		return errors.ES(errors.OpFileIngest, errors.KBlobstore, "no Kusto queue resources are defined, there is no queue to upload to").SetNoRetry()
+	}
+
+	blobName := fmt.Sprintf("%s_%s_%s_%s.gz", filepath.Base(uuid.New().String()), nower(), i.db, i.table)
+
+	// Here's how to upload a blob.
+	blobURL := to.NewBlockBlobURL(blobName)
+
+	gstream := gzip.New()
```
danield137, line 131: shouldn't we make sure we are not double-zipping?

element-of-surprise: I don't think so.  That is kind of an infinite loop.  They can do other compression other than gzip.  

We tell them in the docs specifically not to compress the stream (and to be fair, most people wouldn't do that anyways).

PR #31, opened 2020-06-14
Ingestion from Local File fix for Windows

yogilad:
#### Pull Request Description

IngestFromFile treated local files as blobs on Windows
Add an manual ingestion benchmark tester (requires some more work)

---

#### Future Release Comment

**Fixes:**
- Fix a bug where local files on Windows were treated as blobs causing an ingestion failure


yogilad:
Apologies, This requires some more work.
I'll push this via the ingestion formats change

PR #32, opened 2020-06-15
Support More File Formats

yogilad:
#### Pull Request Description

* Suppert more ingestion formats
* Ingestion works in windows (used to have a problem discerning files from blobs)
* ETE can run with az-cli instead of App credentials (still expects cluster & db in Env) 
* Benchmark tester for ingestion, with memory analysis using pprof

---

#### Future Release Comment
**Breaking Changes:**
- None

**Features:**
- Add support for more file formats to ingest 

**Fixes:**
- Ingestion from local files works in Windows


kusto/ingest/ingest.go
```diff
@@ -119,52 +118,60 @@ func FlushImmediately() FileOption {
...
 	// SCSV is a file containing semicolon ";" separated values.
-	SCSV DataFormat = 8
+	SCSV DataFormat = properties.SCSV
 	// SOHSV is a file containing SOH-separated values(ASCII codepont 1).
-	SOHSV DataFormat = 9
-	// TSV is a file containing table seperated values ("\t").
-	TSV DataFormat = 10
+	SOHSV DataFormat = properties.SOHSV
+	// SINGLEJSON indicates the source containg a single Javascript Object Notation encoded record, newlines are treated as whitespace
+	SINGLEJSON DataFormat = properties.SINGLEJSON
```
kerend, line 154: I would not include it, as it is supported by MULTIJSON format

kusto/ingest/ingest.go
```diff
@@ -119,52 +118,60 @@ func FlushImmediately() FileOption {
...
+	// SSTREAM indicats the source is encoded as a Microsoft Cosmos Structured Streams format
+	SSTREAM DataFormat = properties.SSTREAM
```
kerend, line 156: IIRC, SStream & WCLOGFILE are MS internal, do we want to include them in our SDKs?

yogilad: Hi @vladikbr, 
IIRC, you asked to add all types. 
Are there internal, redundant or soon to be deprecated formats which should not be included?

kusto/ingest/internal/properties/properties.go
```diff
@@ -68,101 +72,140 @@ func (d DataFormat) CamelCase() string {
...
-		return "sohsv", nil
-	case TSV:
-		return "tsv", nil
-	case TXT:
-		return "txt", nil
-	default:
-		return "", fmt.Errorf("EncodingType(%d) was no one we understand", et)
+func dfGetJSONName(et DataFormat) (string, error) {
+	if et > 0 && int(et) < len(dataFormatDescriptionVec) {
+		return dataFormatDescriptionVec[et].JSONName, nil
```
element-of-surprise, line 159: dataFormatDescriptionVec 

That is more like java naming conventions, which is anti-goish.

We don't include types in names, so it is verboten to do:  myMapVar or in this case Vec.

dfDesc or something should be enough.

kusto/ingest/internal/properties/properties.go
```diff
@@ -68,101 +72,140 @@ func (d DataFormat) CamelCase() string {
...
+	// TSVE is a file containing escaped-tab seperated values ("\t").
+	TSVE DataFormat = 15
 	// TXT is a text file with lines deliminated by "\n".
-	TXT DataFormat = 11
+	TXT DataFormat = 16
+	// WCLOGFILE indicates the source is encoded using W3C Extended Log File format
+	WCLOGFILE DataFormat = 17
 )
 
+type dataFormatDescriptior struct {
```
element-of-surprise, line 127: names of all attributes should start lower case: aka camelCaseName, ....

Also, shorten names:

camelName
jsonName
ext
mappingReq

kusto/ingest/internal/properties/properties.go
```diff
@@ -68,101 +72,140 @@ func (d DataFormat) CamelCase() string {
...
+	CamelCaseName    string
+	JSONName         string
+	Extension        string
+	MappingRequired  bool
+	ValidMappingKind bool
+}
+
+var dataFormatDescriptionVec = []dataFormatDescriptior{
```
element-of-surprise, line 135: I'm on the fence on if we should go forward with this, as we are kinda violating the law of least surprise to make it more compact.

This is now getting double use, both for lookups and for extension detection.

But several of these like MultiJson aren't detectable from the file name.  .json is always Json.  We are having to put in comments to explain this.  

I think at least we need to divide the use cases up.  This can stay as is, but formatDetection should go via a new slice:  formatdetect that only includes things we are going to decode.  We should not need comments.

kusto/ingest/internal/properties/properties.go
```diff
@@ -68,101 +72,140 @@ func (d DataFormat) CamelCase() string {
...
-	default:
-		return "", fmt.Errorf("EncodingType(%d) was no one we understand", et)
+	if et > 0 && int(et) < len(dataFormatDescriptionVec) {
+		return dataFormatDescriptionVec[et].CamelCaseName, nil
+	}
+
+	return "", fmt.Errorf("EncodingType(%d) was no one we understand", et)
+}
+
+// DataFormatRequiresMapping indicates whether a data format must be provided with valid mapping file information
```
element-of-surprise, line 173: Move this to a method on DataFormat called RequiresMapping()

kusto/ingest/benchmark/ingest_benchmark.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/Azure/azure-kusto-go/kusto"
+	"github.com/Azure/azure-kusto-go/kusto/ingest"
+	"github.com/Azure/azure-kusto-go/kusto/unsafe"
+	"github.com/Azure/go-autorest/autorest/azure/auth"
+)
+
+//var databaseName = "benchmark_db"
+//var targetCluster = "https://yogiladadx.dev.kusto.windows.net"
+
+var databaseName = "e2e"
```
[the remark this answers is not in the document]
yogilad: Add usage and commandline arguments to provide hard coded options

kusto/ingest/benchmark/ingest_benchmark.go
```diff
@@ -0,0 +1,178 @@
...
+var targetCluster = "https://sdkse2etest.eastus.kusto.windows.net"
+var tableName = "benchmark_table"
+var databaseExists = true
+var artifactFolderPath = "artifacts"
+var csvFilepath = strings.Join([]string{artifactFolderPath, "dataset.csv"}, "/")
+var jsonFilepath = strings.Join([]string{artifactFolderPath, "dataset.json"}, "/")
+var jsonMappingFilePath = strings.Join([]string{artifactFolderPath, "json_mapping.json"}, "/")
+var jsonMappingName = "benchmark_json_mapping"
+
+func main() {
```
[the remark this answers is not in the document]
yogilad: Perhaps 'benchmark' is not a good choice for a name.
The idea of the tester was to inspect memory after n operations, not to time them (though it's quite possible).

I'll have a closer look at the benchmark framework and see if it fits.
in the meantime, perhaps a more appropriate name would suite this runner.
WDYT?

kusto/ingest/benchmark/ingest_benchmark.go
```diff
@@ -0,0 +1,178 @@
...
+	pprofEndpoint := "localhost:6060"
+
+	fmt.Println("Starting memory benchmark")
+	fmt.Println("Please make sure az-cli is installed and 'az login' was called")
```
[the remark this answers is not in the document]
yogilad: Using credentials is cumbersome in my opinion.
Not sure how to test for az existence, but the auth. is explained in the new usage message.

kusto/ingest/benchmark/ingest_benchmark.go
```diff
@@ -0,0 +1,178 @@
...
+	client, err := kusto.New(targetCluster, authorization)
+	if err != nil {
+		fmt.Printf("failed to create a kusto client - %s\n", err.Error())
+		return nil, err
+	}
+
+	return client, nil
+}
+
+func setupBenchmarkDatabase(client *kusto.Client) error {
```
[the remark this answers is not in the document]
yogilad: I think the tool itself is very useful for devs.
I prefer not to cearte a repo for this.
Perhaps if we create a separate folder beside Kusto to keep DevTools?

kusto/ingest/ingest.go
```diff
@@ -119,52 +118,60 @@ func FlushImmediately() FileOption {
 // Not all options can be used in every method.
 type DataFormat = properties.DataFormat
 
+// [Yochai 15.06.2020]
+// John, I want to delete this and use properties.DataFormat insetad, but I get the feeling this will cause a breaking change.
+// WDYT?
+
```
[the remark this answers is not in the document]
yogilad: This basically covers it, unless we actually add a new format to Kusto itself.
This can be kept as is.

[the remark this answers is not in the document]
yogilad: Comments (mine and yours) removed

kusto/ingest/internal/properties/properties.go
```diff
@@ -68,101 +72,140 @@ func (d DataFormat) CamelCase() string {
...
-		return "scsv", nil
-	case SOHSV:
-		return "sohsv", nil
-	case TSV:
-		return "tsv", nil
-	case TXT:
-		return "txt", nil
-	default:
-		return "", fmt.Errorf("EncodingType(%d) was no one we understand", et)
+func dfGetJSONName(et DataFormat) (string, error) {
```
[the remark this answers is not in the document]
yogilad: Should I change MarshalJSON as well?

kusto/ingest/internal/properties/properties.go
yogilad: dataFormatDescriptions

kusto/ingest/internal/properties/properties.go
yogilad: Renamed to DetectableExtension to specify the use.
Ambiguous formats are now "" empty strings so order is not important
Comments removed

kusto/ingest/internal/filesystem/filesystem.go
```diff
@@ -362,3 +325,23 @@ func CompressionDiscovery(fName string) properties.CompressionType {
 	}
 	return properties.CTNone
 }
+
+var (
+	// gExtractURIProtocol is created outside the fucntion inorder to pay once for regexp creation.
```
element-of-surprise, line 330: Not a fan of regexes, usually lead to something bad.  I'm an avoid if can.  Since we don't want to use a lexer/parser here, let's use url to our advantage:

https://play.golang.org/p/QplM9oEEGRm

```go
u, err := url.Parse(s)
if err != nil {
  return true
}
switch u.Scheme {
case "http", "https":
  return false
}
return true
```

kusto/ingest/internal/filesystem/filesystem_test.go
```diff
@@ -40,9 +40,9 @@ func TestFormatDiscovery(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		got := FormatDiscovery(test.input)
+		got := properties.DataFormatDiscovery(test.input)
 		if got != test.want {
-			t.Errorf("TestFormatDiscovery(%s): got %s, want %s", test.input, got, test.want)
+			t.Errorf("TestFormatDiscovery(%s): got '%s', want '%s'", test.input, got, test.want)
```
element-of-surprise, line 45: Don't add the quotes.

If you absolutely thing you need them, use %q.

kusto/ingest/internal/properties/properties.go
```diff
@@ -52,117 +55,161 @@ const (
...
 	}
+
 	return ext
 }
 
 // CamelCase returns the CamelCase version. This is for internal use, do not use.
 // This can be removed in future versions.
 func (d DataFormat) CamelCase() string {
-	cc, err := dfToCamel(d)
+	cc, err := d.camelCase()
```
element-of-surprise, line 139: Collapse camelCase() into CamelCase()

kusto/ingest/internal/properties/properties.go
```diff
@@ -52,117 +55,161 @@ const (
...
+	{"Tsv", "tsv", ".tsv", false, false},
+	{"Tsve", "tsve", ".tsve", false, false},
+	{"Txt", "txt", ".txt", false, false},
+	{"EcLogFile", "eclogfile", ".eclogfile", false, false},
+}
+
 // String implements fmt.Stringer.
 func (d DataFormat) String() string {
-	ext, err := dfToExt(d)
+	ext, err := d.jsonName()
```
element-of-surprise, line 128: Unless .jsonName() is getting used somewhere else where you are doing error detection, collapse jsonName() into String()

kusto/ingest/benchmark/ingest_benchmark.go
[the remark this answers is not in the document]
yogilad: I'll just keep a local copy on my Dev Station

kusto/ingest/ingest.go
```diff
@@ -119,52 +118,53 @@ func FlushImmediately() FileOption {
...
-	// CSV indicates the source is encoded in comma seperated values.
-	CSV DataFormat = 1
-	// JSON indicates the source is encoded in Javscript Object Notation.
-	JSON DataFormat = 2
+	DFUnknown DataFormat = properties.DFUnknown
 	// AVRO indicates the source is encoded in Apache Avro format.
-	AVRO DataFormat = 3
-	// Parquet indicates the source is encoded in Apache Parquet format.
-	Parquet DataFormat = 4
+	AVRO DataFormat = properties.AVRO
```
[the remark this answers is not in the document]
yogilad: Are you referring to the comments?
I just repeated the exiting comment comments "format".
I can delete all if you think this is redundent

kusto/ingest/ingest.go
```diff
@@ -119,52 +118,53 @@ func FlushImmediately() FileOption {
...
-
 // IngestionMapping provides runtime mapping of the data being imported to the fields in the table.
 // "ref" will be JSON encoded, so it can be any type that can be JSON marshalled. If you pass a string
 // or []byte, it will be interpreted as already being JSON encoded.
 // mappingKind can only be: CSV, JSON, AVRO, Parquet or ORC.
 func IngestionMapping(mapping interface{}, mappingKind DataFormat) FileOption {
 	return propertyOption(
 		func(p *properties.All) error {
-			if !validMappingKind[mappingKind] {
+			if !mappingKind.IsValidMappingKind() {
```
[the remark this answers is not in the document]
yogilad: That would have been true if mappingKind was a type of it's own, but it is DataFormat type.

kusto/ingest/ingest.go
```diff
@@ -356,11 +356,11 @@ func (i *Ingestion) FromFile(ctx context.Context, fPath string, options ...FileO
 		}
 	}
 
-	if _, err := url.ParseRequestURI(fPath); err == nil {
-		return i.fs.Blob(ctx, fPath, 0, props)
+	if filesystem.IsFileSystem(fPath) {
```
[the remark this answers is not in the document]
yogilad: I considered this as well, but a shared file system path is not local, and yet it can be opened via a filesystem API.
E.g. \\host\drive\folder\path if not local, but it is a file system path

kusto/ingest/internal/filesystem/filesystem.go
yogilad: Using URL was what caused the problem in the first place.
Also, there are other schemes such as 'ftp' which are not file system, so your suggested implementation is inaccurate.

kusto/ingest/internal/filesystem/filesystem_test.go
yogilad: Ok.
I want to make sure that if by any change we get an empty string, it will be visible in the error

kusto/ingest/ingest.go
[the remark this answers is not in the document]
element-of-surprise: No, I saying that if, for example what you did on the last commit (where you removed SingleJSON), that we should only have formats that aren't currently deprecated and no internal formats.

kusto/ingest/ingest.go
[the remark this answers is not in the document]
element-of-surprise: I think that is probably getting too literal.  For our purposes, it is a local path because it is hosted locally as a drive, regardless if the mount id a local disk, remote array, NFS or CIFS share, .....  FileSystem, however, is too generic.  Everything is a filesystem, including blob store.  

LocalPath indicates it is mounted on the user system and in that case we use our selected method.

kusto/ingest/internal/filesystem/filesystem.go
element-of-surprise: You are right, but this leaves me liking this even less.

I don't think we handle file:// and I'm sure we don't handle "ftp://".  Its either blob or local file.  This of course is going to error somewhere farther down, but I'm not a huge fan of that.  

Filesystems, so many corner cases when you mix them.  Blah.

I don't want to enumerate every possible scheme, and because windows does c:\, it shows up as a url scheme.  
I also really, really don't want to do a lex/parser, cause that is a rabbit hole for little gain.

Maybe we just need to rethink this:  

```go
func IsLocalPath(s string) (bool, error) {
  u, err := url.Parse(s)
  if err != nil {
    return true
  }
  switch u.Scheme {
  // With this we know it SHOULD be a blobstore path.  It might not be, but I think that is a fine assumption to make.
  case "http", "https":
    return false, nil
  }

  // By this point, we know its not blobstore, so it needs to be something that gets resolved to a file.  
  // So we are going to Stat() the file and see if it exists and is not a directory.  
  // In your tests, this would fail "file://" which we don't support.  Also, because of this method, your tests
  // are going to be broken.  Can't test unix paths on windows, can't test windows path on unix.  
  // Best thing we can probably do is do tests with build tags.  Again, fileystems, blah....
  if stat, err := os.Stat(s); err != nil {
    return false, fmt.Errorf("is it a valid local file path(could not stat file) and was not a valid blob path")
  }
  if stat.IsDir() {
    return false, fmt.Errorf("path is a local directory and not a valid file")
  }
  return true, nil
}
```

kusto/ingest/internal/properties/properties.go
element-of-surprise: Don't know that I'm a fan, but it might be that I'm getting to picky here.  Let's let this go forward.

kusto/ingest/ingest.go
```diff
@@ -124,47 +123,49 @@ type DataFormat = properties.DataFormat
...
 	// SCSV is a file containing semicolon ";" separated values.
-	SCSV DataFormat = 8
+	SCSV DataFormat = properties.SCSV
 	// SOHSV is a file containing SOH-separated values(ASCII codepont 1).
-	SOHSV DataFormat = 9
-	// TSV is a file containing table seperated values ("\t").
-	TSV DataFormat = 10
+	SOHSV DataFormat = properties.SOHSV
+	// SStream indicats the source is encoded as a Microsoft Cosmos Structured Streams format
+	SStream DataFormat = properties.SStream
```
element-of-surprise, line 150: Please remove as this is MS internal according to comments

kusto/ingest/ingest.go
```diff
@@ -124,47 +123,49 @@ type DataFormat = properties.DataFormat
...
+	// TSV is a file containing tab seperated values ("\t").
+	TSV DataFormat = properties.TSV
+	// TSVE is a file containing escaped-tab seperated values ("\t").
+	TSVE DataFormat = properties.TSVE
 	// TXT is a text file with lines deliminated by "\n".
-	TXT DataFormat = 11
+	TXT DataFormat = properties.TXT
+	// WCLogFile indicates the source is encoded using W3C Extended Log File format
+	W3CLogFile DataFormat = properties.W3CLogFile
```
element-of-surprise, line 158: Please remove as this is MSInternal per comments

kusto/ingest/internal/filesystem/filesystem_test.go
```diff
@@ -212,3 +213,31 @@ func TestLocalToBlob(t *testing.T) {
 		}
 	}
 }
+
+func TestIsLocalFileSystem(t *testing.T) {
```
element-of-surprise, line 217: Rewrote tests are in the same snippet:
https://play.golang.org/p/129aZ1p_Zyu

kusto/ingest/internal/properties/properties.go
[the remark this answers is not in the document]
element-of-surprise: I'm not sure what you would be changing as it is already a method on the type.  The changes you made seem to be correct to me.  

If there is something I'm missing, please let me know.

kusto/ingest/ingest.go
yogilad: As discussed, keeping all but Single Json

kusto/ingest/internal/properties/properties.go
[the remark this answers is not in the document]
yogilad: I was wondering if this is some implicit contract with the Json serializer that I break.
But all tests pass, so...

kusto/ingest/ingest.go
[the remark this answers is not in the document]
yogilad: Oh, ok.
We had a length internal conversation on this and decided to keep all except SingleJson

kusto/ingest/ingest.go
yogilad: We discussed this internally. 
Although Go is not commonly used in MS, the SDK serves both MS and non MS develpments and therefore should support all formats.

PR #34, opened 2020-08-24
Add Ingestion Status Reporting via Azure Tables

yogilad:
#### Pull Request Description

Support ingestion status reporting via azure tables.

---

#### Future Release Comment
**Breaking Changes:**
- Support ingestion status reporting via azure tables.
 `ingest.FromFile` and `ingest.FromReader `now return a channel where the status can be awaited on.


kusto/ingest/result.go
```diff
@@ -0,0 +1,149 @@
...
+	} else {
+		r.tableClient = client
+	}
+
+	return r
+}
+
+// Wait returns a channel that can be checked for ingestion results.
+// In order to check actual status please use the IngestionStatus option when ingesting data.
+func (r *Result) Wait(ctx context.Context) chan StatusRecord {
```
element-of-surprise, line 98: So I want to go back to this.  I think StatusRecord should probably be internal/  and this should probably return error.

That error can be a StatusRecord, if status record implements the "Error() string" method.  Then we can return a mix of errors.  There is no use for StatusRecord itself.  

There are only a few states that the user will receive, which basically say: pass/fail/partial.  That is all the user is going to care about + the text message.

And if we changed putErr() to store the actual error when doing client errors, you could return a client specific error so they could check if they should retry easily.  Or we could return (*Result, error) where error is always client errors (honestly I like this better).  

```go
type ErrClient struct {
  Msg string
}
func (e ErrClient) Error() string {
  return e.Msg
}
```

But I want to get rid of access to StatusRecord and probably do a statusRecord which implements Error(). 

Also, going to put nots on StatusRecord below.

kusto/ingest/result.go
```diff
@@ -0,0 +1,149 @@
...
+
+	return ch
+}
+
+func (r *Result) poll(ctx context.Context) {
+	// create a table client
+	if r.tableClient != nil {
+		// Create a ticker to poll the table in 10 second intervals.
+		ticker := time.NewTicker(10 * time.Second)
+
```
element-of-surprise, line 119: This is a common source of goroutine leaks.  NewTicket() creates a goroutine that you never release.  So each call here will leak memory/cpu.  You should do a defer ticker.Stop().

However, I don't think you want a ticker as it is going to behave counter to what you want.  A ticker delivers ticks at set intervals.  But you are doing something that may take longer than the tick.  In that case the ticker will immediately hit and will most likely there will be no time breaks.

You probably want to use NewTimer() like so:

```go
t := time.NewTimer(10 * time.Second)
defer t.Stop()

for {
  t.Reset(10 * time.Second) // Timer only fires once unless reset. This guarantees 10 seconds between calls.
  ...
}
```

kusto/ingest/status.go
```diff
@@ -0,0 +1,291 @@
...
+	r.Table = props.Ingestion.TableName
+	r.UpdatedOn = time.Now()
+
+	if props.Ingestion.BlobPath != "" && r.IngestionSourcePath == "Undefined" {
+		r.IngestionSourcePath = props.Ingestion.BlobPath
+	}
+}
+
+// FromMap converts an ingestion status record to a key value map.
+func (r *StatusRecord) FromMap(data map[string]interface{}) {
```
element-of-surprise, line 145: So, this CL seems incomplete, which I imagine means you checked in some code already.  Though it could be I don't do a lot of github reviews.

I don't see GetEntityReference here, which would probably confirm my suspicions.  But from what I'm seeing, FromMap() is probably converting a json.Unmarshal() into a map[string]interface{}.  

I don't think we should probably be doing that, instead we should be doing a json.Unmarshal directly into a StatusRecord.  

json decoding not being anyone's favorite, my guess is that there were some problems and that lead to this conversion into map[string]interface{} and then into the concrete type.

If that isn't wrong, then we should talk about what the problems were and eliminate all this map[string]interface{} code.

kusto/ingest/status.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	if data["OriginatesFromUpdatePolicy"] != nil {
+		r.OriginatesFromUpdatePolicy = data["OriginatesFromUpdatePolicy"].(bool)
+	}
+
+}
+
+// ToMap converts an ingestion status record to a key value map.
+func (r *StatusRecord) ToMap() map[string]interface{} {
```
element-of-surprise, line 209: Same as above.

kusto/ingest/status.go
```diff
@@ -0,0 +1,291 @@
...
+	data["IngestionSourcePath"] = r.IngestionSourcePath
+	data["Database"] = r.Database
+	data["Table"] = r.Table
+	data["UpdatedOn"] = r.UpdatedOn.Format(time.RFC3339Nano)
+
+	return data
+}
+
+// String converts an ingestion status record a printable  string.
+func (r *StatusRecord) String() string {
```
element-of-surprise, line 226: I'd probably just do:

return pretty.Sprint(r)

Think that will not only get you a better result, but if we add fields we don't have to go back and update this.

kusto/ingest/status.go
```diff
@@ -0,0 +1,291 @@
...
+		r.UpdatedOn,
+		r.OperationID,
+		r.ActivityID,
+		r.OriginatesFromUpdatePolicy,
+		r.Details)
+
+	return str
+}
+
+// ToError converts an ingestion status to an error if failed or partially succeeded, or nil if succeeded.
```
element-of-surprise, line 245: As I talk about above, I think we should be returning error and not StatusRecord.  When we fail, we should return StatusRecord with a new method:

```go
func (r *StatusRecord) Error() string {
  return r.String()
}

File: kusto/ingest/status.go [hunk 7]
```diff
@@ -0,0 +1,291 @@
...
+	data["IngestionSourceId"] = r.IngestionSourceID
+	data["IngestionSourcePath"] = r.IngestionSourcePath
+	data["Database"] = r.Database
+	data["Table"] = r.Table
+	data["UpdatedOn"] = r.UpdatedOn.Format(time.RFC3339Nano)
+
+	return data
+}
+
+// String converts an ingestion status record a printable  string.
```
element-of-surprise, line 225: This comments should say:

// String implements fmt.Stringer.

kusto/ingest/status.go
```diff
@@ -0,0 +1,291 @@
...
+		return nil
+
+	case PartiallySucceeded:
+		return fmt.Errorf("Ingestion succeeded partially\n" + r.String())
+	}
+
+	return fmt.Errorf("Ingestion Failed\n" + r.String())
+}
+
+func getTimeFromInterface(x interface{}) (time.Time, error) {
```
element-of-surprise, line 259: If we get rid of this map stuff, this shouldn't be required.

element-of-surprise (changes requested):
I didn't get to the tests, because there are some fundamental changes I want to make here.  Comments in and I expect you'll want to meet about these.

go.mod
```diff
@@ -3,12 +3,14 @@ module github.com/Azure/azure-kusto-go
...
 require (
+	github.com/Azure/azure-sdk-for-go v44.1.0+incompatible
 	github.com/Azure/azure-storage-blob-go v0.8.0
 	github.com/Azure/azure-storage-queue-go v0.0.0-20191125232315-636801874cdd
 	github.com/Azure/go-autorest/autorest v0.10.0
 	github.com/Azure/go-autorest/autorest/azure/auth v0.4.2
 	github.com/google/uuid v1.1.1
 	github.com/kylelemons/godebug v1.1.0
 	github.com/pkg/errors v0.9.1 // indirect
+	github.com/satori/go.uuid v1.2.0
```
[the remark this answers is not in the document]
yogilad: It's forced upon us by Storage SDK, since I have to explicitly convert between it and google uuid

go.sum
```diff
@@ -49,6 +52,8 @@ github.com/mitchellh/go-homedir v1.1.0 h1:lukF9ziXFxDFPkA1vsr5zpc1XuPDn/wFntq5mG
 github.com/mitchellh/go-homedir v1.1.0/go.mod h1:SfyaCUpYCn1Vlf4IUYiD9fPX4A5wJrkLzIz1N1q0pr0=
 github.com/pkg/errors v0.9.1 h1:FEBLx1zS214owpjy7qsBeixbURkuhQAwrK5UwLGTwt4=
 github.com/pkg/errors v0.9.1/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINEl0=
+github.com/satori/go.uuid v1.2.0 h1:0uYX9dsZ2yD7q2RtLRtPSdGDWzjeM3TbMJP9utgA0ww=
+github.com/satori/go.uuid v1.2.0/go.mod h1:dA0hQrYB0VpLJoorglMZABFdXlWrHn1NEOzdhQKdks0=
```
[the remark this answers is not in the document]
yogilad: See above comment

kusto/ingest/ingest.go
```diff
@@ -261,6 +261,17 @@ func IfNotExists(ingestByTag string) FileOption {
 	)
 }
 
+// ReportResultToTable option requests that the ingestion status will be tracked in an Azure table
```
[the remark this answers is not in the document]
yogilad: The next step would be to add queue status reporting support.
Perhaps I'll change this to take in options now.
I''ll add the warning

kusto/ingest/ingest.go
```diff
@@ -331,82 +342,140 @@ type mapEntry struct {
...
 	for _, o := range options {
 		if propOpt, ok := o.(propertyOption); ok {
 			if err := propOpt(&props); err != nil {
-				return err
+				return result.putErr(err)
 			}
 		}
 	}
 
+	if props.Ingestion.ReportMethod == properties.ReportStatusToTable || props.Ingestion.ReportMethod == properties.ReportStatusToQueueAndTable {
```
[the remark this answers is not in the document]
yogilad: I disagree. 
There are no present or future cases other then the coded one.

kusto/ingest/ingest.go
```diff
@@ -331,82 +342,140 @@ type mapEntry struct {
...
 	for _, o := range options {
 		if propOpt, ok := o.(propertyOption); ok {
 			if err := propOpt(&props); err != nil {
-				return err
+				return result.putErr(err)
+			}
+		}
+	}
+
+	if props.Ingestion.ReportMethod == properties.ReportStatusToTable || props.Ingestion.ReportMethod == properties.ReportStatusToQueueAndTable {
```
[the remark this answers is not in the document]
yogilad: I agree, but it was there to begin with, and this PR is too complicated as is.
I have a few small changes I reserve for a later PR.
I'll add it there

kusto/ingest/internal/filesystem/filesystem.go
```diff
@@ -107,24 +107,24 @@ func (i *Ingestion) Local(ctx context.Context, from string, props properties.All
 }
 
 // Reader uploads a file via an io.Reader.
-func (i *Ingestion) Reader(ctx context.Context, reader io.Reader, props properties.All) error {
+func (i *Ingestion) Reader(ctx context.Context, reader io.Reader, props properties.All) (string, error) {
```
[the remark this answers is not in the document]
yogilad: I'm doing that so that the ingestion status record (local and remote) will contain a reference to the blob URI that has succeeded / failed

go.mod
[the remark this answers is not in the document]
element-of-surprise: Sounds like I need to go and update storage to use the proper one. <sigh>

kusto/ingest/status.go
element-of-surprise: Oh, ......

Well that is just horrible.  As a note, if you ever see map[string]interface{}, it is almost always someone being lazy about a JSON conversion.  

This is absolutely terrible that they are giving this back to you instead of a struct.

So ignore my comment.

kusto/ingest/result.go
element-of-surprise: No, because then the person could tease the StatusRecord back by doing:

err := blah.Wait()
if sr, ok := err.(StatusRecord); ok {
  // Do stuff with StatusRecord
}

So StatusRecord would also need to become statusRecord + ToError() -> Error(), preventing them from doing that.  Or it would need to be in an internal/ directory.

You can leak a private type with public methods which gives them access to public methods but not the type.

But I'm not in love with this.  

Here's what I think we should do:

```go
FromReader(ctx context.Context, reader io.Reader, options ...FileOption) *Result
```
should become
```go
FromReader(ctx context.Context, reader io.Reader, options ...FileOption) (*Result, error)
```

The error would be a clientError from status.go

This differentiates clearly our errors that aren't directly ingestion data errors.

Status records should move to internal/ or become statusRecord

You can keep ToError() and just return that in *Result.Wait().

If later we want to distinguish between these error types (and I can't see why we would programmically care), we can still do that.

kusto/ingest/status.go
```diff
@@ -0,0 +1,277 @@
...
+
+	if props.Ingestion.BlobPath != "" && r.IngestionSourcePath == "Undefined" {
+		r.IngestionSourcePath = props.Ingestion.BlobPath
+	}
+}
+
+// FromMap converts an ingestion status record to a key value map.
+func (r *StatusRecord) FromMap(data map[string]interface{}) {
+	if data["Status"] != nil {
+		r.Status = StatusCode(data["Status"].(string))
```
element-of-surprise, line 148: I should make a note here to make you aware.  If they ever return a non-string for "Status", this will panic.
I think that is probably fine.  But if you want to be sure that this won't crash, you need to do:

```go
if v := data["Status"]; v != nil {
  s, ok := v.(string)
  if !ok {
    log.Printf("bug: backend service changed StatusRecord's "Status" type from string to %T", v)
  }
  r.Status = StatusCode(s)
}

Also, if you want to make this a little faster and a little cleaner, here and below you can do:

```go
if v := data["Status"]; v != nil {
  r.Status = StatusCode(v.(string))
}
```
This prevents having to do two hash lookups.

kusto/ingest/ingest.go
```diff
@@ -331,82 +342,140 @@ type mapEntry struct {
...
 		}
 	}
 
+	if props.Ingestion.ReportMethod == properties.ReportStatusToTable || props.Ingestion.ReportMethod == properties.ReportStatusToQueueAndTable {
+		if props.Source.ID == uuid.Nil {
+			props.Source.ID = uuid.New()
+		}
+
+		resources, err := manager.Resources()
+		if err == nil {
```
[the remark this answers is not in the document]
yogilad: Hmm...
Would you want to fail the operation or disable status reporting?
I think the former, but this feels harsh

kusto/ingest/result.go
element-of-surprise: Thinking of our goals from today, I think what I'm proposing below will meet these:
- Hide the implementation (StatusRecord) from the user
- Allow the user to know if they can retry an error
- The error must have diagnostic information from StatusRecord for support calls
- Allow the possibility, in the future, of exposing StatusRecord via Wait(), if required

## Proposed changes

### StatusRecord to become private
StatusRecord -> statusRecord

### Make statusRecord implement the built in error type

statusRecord.Error() ->
```go
// Error implements error.Error().
func (r *statusRecord) Error() string {
	return r.String()
}
```
Notes:
	- We will only return statusRecord as an error if the status was not Succeeded.  Any other status is a failure and they will never see status like Queued.
	- When there is an error, it will give them the string output of the StatusRecord for support calls.
	- This always gives us the option, at a later date, to make StatusRecord public and expose it to the user

### Change Result.Wait to return a chan error
Result.Wait() ->
```go
func (r *Result) Wait(ctx context.Context) chan error {
...
}
```
Notes:
	- If there is an error, you return the statusRecord, which now implements error
	- Otherwise you just return nil


### Provide the ability to know if an error can be retried
```go
func CanRetry(e error) bool{
	if sr, ok := e.(statusRecord); ok {
	    switch sr.FailureStatus{
	    case Unknown, Permanent:
		    return false
	    }
	    return true
        }
        if errors.Is(e, context.Canceled || errors.Is(e, context.DeadlineExceeded) {
            return true
        }
        return false
}
```

## Final notes
Notes: 
	- FailureStatusCode would become failureStatusCode and all instances would be private (we can always change that later)
	- StatusCode would become private and all instances (we can always change that later)
	- ToMap() should become toMap() (I didn't see any reason for it to be public, but I could be wrong)
	- FromMap() should become fromMap() (same)

kusto/ingest/ingest.go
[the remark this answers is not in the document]
element-of-surprise: Fail the operation.

I generally follow the law of least surprise.  Asking for status reporting but not getting it, but the operation happens and may or may not fail seems like a lot of surprises.

kusto/ingest/result.go
```diff
@@ -0,0 +1,172 @@
...
+// newResult creates an initial ingestion status record.
+func newResult() *Result {
+	ret := &Result{}
+
+	ret.record = newStatusRecord()
+	return ret
+}
+
+// putProps sets the record to a failure state and adds the error to the record details.
+func (r *Result) putProps(props properties.All) *Result {
```
element-of-surprise, line 30: Here and below, I'm not sure why the methods are returning the type.  I looked at a couple places where you use putQueue() and putProps(), but I don't see you using that.  

Usually this is only done if you are going to do some type of chaining, like:

```go
  <-result.putQueue(mgr).putProps(props).Wait()
```

kusto/ingest/result.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		r.record.Status = StatusRetrievalFailed
+		r.record.FailureStatus = Permanent
+		r.record.Details = "Failed Creating a Status Table client: " + err.Error()
+		return r
+	}
+
+	// Write initial record
+	r.record.Status = Pending
+	recordMap := r.record.ToMap()
```
element-of-surprise, line 72: Remove this line and just integrate it into the next:

```go
err = client.Write(r.record.IngestionSourceID.String(), r.record.ToMap())
```

kusto/ingest/result.go
```diff
@@ -0,0 +1,172 @@
...
+
+// Wait returns a channel that can be checked for ingestion results.
+// In order to check actual status please use the IngestionStatus option when ingesting data.
+func (r *Result) Wait(ctx context.Context) chan error {
+	ch := make(chan error, 1)
+
+	go func() {
+		defer close(ch)
+
+		if !r.record.Status.IsFinal() && r.reportToTable == true {
```
element-of-surprise, line 93: r.reportToTable == true can just be r.reportToTable

However, as above statement, I think we should get rid of this if statement.

kusto/ingest/result.go
```diff
@@ -0,0 +1,172 @@
...
+			r.poll(ctx)
+		}
+
+		if !r.record.Status.IsSuccess() {
```
element-of-surprise, line 97: I think if you want to make this slightly cleaner, you could have poll() return the success status.  

Then the entire goroutine would be:

```go
go func() {
  defer close(ch)
  if !r.poll() {
    ch <-r.record
  }
}(()
```

or you could just have poll do it all:

```go
ch := make(chan error, 1)
// If status reporting hasn't been turned on, we always report a nil error and immediately return.
if !r.reportToTable {
  close(ch)
  return ch
}
go r.poll(ctx, ch)
return ch
```

kusto/ingest/result.go
```diff
@@ -0,0 +1,172 @@
...
+				r.record.Status = StatusRetrievalCanceled
+				r.record.FailureStatus = Transient
+				return
+
+			// Whenever the ticker fires.
+			case <-timer.C:
+				// read the current state
+				smap, err := r.tableClient.Read(r.record.IngestionSourceID.String())
+				if err != nil {
+					// Read failure
```
element-of-surprise, line 128: And here.

kusto/ingest/result.go
```diff
@@ -0,0 +1,172 @@
...
+				}
+
+				timer.Reset(pollInterval)
+			}
+		}
+	}
+}
+
+// GetIngestionStatus extracts the ingestion status code from an ingestion error
+func GetIngestionStatus(err error) (StatusCode, error) {
```
element-of-surprise, line 148: I think we should combine this and GetIngestionFailureStatus() and remove Get*, and drop the extra error, aka:

```go
// IngestionStatus retrieves the StatusCode and FailureStatusCode if the error contains them.  It will return
// UnknownStatus and UnknownFailure if the error does not contain them.
fun IngestionStatus(err error) (StatusCode, FailureStatusCode) {
        if s, ok := err.(statusRecord); ok {
		return s.Status, s.FailureStatus
	}
        return UnknownStatus, UnknownFailure
}
```

This change would require adding: UnknownStatus and changing Unknown to UnknownFailure.

kusto/ingest/status.go
```diff
@@ -0,0 +1,301 @@
...
+	return i != Pending
+}
+
+// IsSuccess returns true if the status code is a final successfull status code
+func (i StatusCode) IsSuccess() bool {
+	switch i {
+	case Succeeded:
+		return true
+
+	case Queued:
```
element-of-surprise, line 55: I'm not sure why Queued is considered succeeded, unless Queued is guaranteed to always complete with a success status.  That seems hard to guarantee that the backend wouldn't change that behavior.  

Also, this can be made more succinct:

```go
switch i {
case Succeeded, Queued:
  return true
}
return false

File: kusto/ingest/status.go [hunk 24]
```diff
@@ -0,0 +1,301 @@
...
+	Permanent FailureStatusCode = "Permanent"
+	// Transient represnet a retryable failure state
+	Transient FailureStatusCode = "Transient"
+	// Exhausted represents a retryable failure that has exhusted all retry attempts
+	Exhausted FailureStatusCode = "Exhausted"
+)
+
+// IsRetryable indicates whether there's any merit in retying ingestion
+func (i FailureStatusCode) IsRetryable() bool {
+	switch i {
```
element-of-surprise, line 79:
```go
switch i {
case Transient, Exhausted:
  return true
}
return false

File: kusto/ingest/status.go [hunk 25]
```diff
@@ -0,0 +1,301 @@
...
+		return true
+
+	default:
+		return false
+	}
+}
+
+// statusRecord is a record containing information regarding the status of an ingation command
+type statusRecord struct {
+	// Status - The ingestion status returned from the service. Status remains 'Pending' during the ingestion process and
```
element-of-surprise, line 93: Comment must be complete sentences.  So you cannot do // Status -

Must be of form:

// Status is the ingestion statues returned from the service. ...

kusto/ingest/status.go
```diff
@@ -0,0 +1,301 @@
...
+	// will always be 'Queued' and the caller needs to query the reports queues for ingestion status, as configured. To query statuses that were
+	// reported to queue, see: <see href="https://docs.microsoft.com/en-us/azure/kusto/api/netfx/kusto-ingest-client-status#ingestion-status-in-azure-queue"/>.
+	// When <see cref="IngestionReportMethod"/> is set to 'Table', call <see cref="IKustoIngestionResult.GetIngestionStatusBySourceId"/> or
+	// <see cref="IKustoIngestionResult.GetIngestionStatusCollection"/> to retrieve the most recent ingestion status.
+	Status StatusCode
+
+	// IngestionSourceID - A unique identifier representing the ingested source. It can be supplied during the ingestion execution.
+	IngestionSourceID uuid.UUID
+
+	// The URI of the blob, potentially including the secret needed to access
```
element-of-surprise, line 104: Must start with the name of the variable, IngestionSourcePath and complete sentences.

kusto/ingest/status.go
```diff
@@ -0,0 +1,301 @@
...
+	// the blob. This can be a filesystem URI (on-premises deployments only),
+	// or an Azure Blob Storage URI (including a SAS key or a semicolon followed
+	// by the account key).
+	IngestionSourcePath string
+
+	// Database - The name of the database holding the target table.
```
element-of-surprise, line 110: All variables here need the comments updated to Go spec.

kusto/ingest/status.go
```diff
@@ -0,0 +1,301 @@
...
+	FailureStatus FailureStatusCode
+
+	// Details - In case of a failure, indicates the failure's details.
+	Details string
+
+	// OriginatesFromUpdatePolicy - In case of a failure, indicates whether or not the failure originated from an Update Policy.
+	OriginatesFromUpdatePolicy bool
+}
+
+// newStatusRecord creates a new record initialized with defaults and user provided data.
```
element-of-surprise, line 139: Comment is not actually no longer this definition.  There is no user provided data here.

kusto/ingest/status.go
```diff
@@ -0,0 +1,301 @@
...
+func newStatusRecord() statusRecord {
+	rec := statusRecord{
+		Status:                     Failed,
+		IngestionSourceID:          uuid.Nil,
+		IngestionSourcePath:        "Undefined",
```
element-of-surprise, line 144: Make "Undefined" a constant so you can use it here and below in other functions like  FromProps().  This prevents misspelling mistakes now and the future.

kusto/ingest/status.go
```diff
@@ -0,0 +1,301 @@
...
+		}
+
+		return uid, err
+
+	default:
+		return uuid.Nil, fmt.Errorf("getGoogleUUIDFromInterface: Unexpected format %T", x)
+	}
+}
+
+func safeSetString(data map[string]interface{}, key string, target *string) {
```
element-of-surprise, line 295: I think we should change how we are doing this to:

```go
func extractKey(m map[string]interface{}, key string) string {
  if v := data[key]; v != nil {
    if s, ok := v.(string); ok {
      return s
    }
  }
  return ""
}
```

Then above we do:

```go
r.IngestionSourcePath = extractKey(data, "IngestionSourcePath")
```
This keeps the operations as string copies, no pointer conversions and prevents any possibility of escape analysis that could have us putting things on the heap.

element-of-surprise (changes requested):

kusto/test/etoe/ingestion_status_e2e_test.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+func TestIgestionFromFileWithStatusReportingQueued(t *testing.T) {
+	err := initOnce()
+	if err != nil {
+		t.Skipf("Skipping tests: %s", err)
+		return
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
element-of-surprise, line 85: Same

kusto/test/etoe/ingestion_status_e2e_test.go
```diff
@@ -0,0 +1,395 @@
+package etoe
```
element-of-surprise, line 1: This is a general comment:

I'm going to let this file go through, however......

Really, in the future this file should be combined with etoe_test.  Those tests should become table driven tests, where one of the tests simply passes the ReportResultToTable() option and the tests code knows about what extra things to check.

kusto/test/etoe/ingestion_status_e2e_test.go
```diff
@@ -0,0 +1,395 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	ingestor, err := ingest.New(client, testConfig.Database, tableName)
+	if err != nil {
+		panic(err)
+	}
+
+	count := 5
```
element-of-surprise, line 99: You can do this in this manner, but it feels odd.  You almost never use arrays in Go.  And you usually only use C style loops for counting usually once for a set of activities.

This is more in the Go fashion.
```go
const numTests = 5

results := make([]*Result, 0, numTests)

for  i := 0; i < numTests, i++ {
  res, err := ingestor.FromFile(ctx, csvFile, ingest.ReportResultToTable())
  if err != nil {
    t.Fatalf("TestIgestionFromFileWithStatusReportingQueued: got err == %q, want err == nil", err)
  }
  results = append(results, res)
}

for i, res := range results {
  if err := res.Wait(ctx); err != nil {
    t.Errorf("TestIgestionFromFileWithStatusReportingQueued(ingestion %d): got err == %q, want err == nil", i, err)
  } 
}
```
This still executes in parallel, eliminates the channels array and errors array, documents the errors better, and has a more Go feel.

kusto/test/etoe/ingestion_status_e2e_test.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+func TestIgestionFromReaderWithStatusReportingQueued(t *testing.T) {
+	err := initOnce()
+	if err != nil {
+		t.Skipf("Skipping tests: %s", err)
+		return
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
element-of-surprise, line 130: Same thing.

kusto/test/etoe/ingestion_status_e2e_test.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+func TestIgestionFromFileWithoutStatusReporting(t *testing.T) {
+	err := initOnce()
+	if err != nil {
+		t.Skipf("Skipping tests: %s", err)
+		return
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
element-of-surprise, line 182: Same

kusto/test/etoe/ingestion_status_e2e_test.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+func TestIgestionFromReaderWithoutStatusReporting(t *testing.T) {
+	err := initOnce()
+	if err != nil {
+		t.Skipf("Skipping tests: %s", err)
+		return
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
element-of-surprise, line 214: Same and below

kusto/test/etoe/ingestion_status_e2e_test.go
```diff
@@ -0,0 +1,395 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	ingestor, err := ingest.New(client, testConfig.Database, tableName)
+	if err != nil {
+		panic(err)
+	}
+
+	// Once without status reporting
```
element-of-surprise, line 271: Make this table driven, your doing double tests of the same thing with an option.

kusto/test/etoe/ingestion_status_e2e_test.go
```diff
@@ -0,0 +1,395 @@
...
+	ingestor, err := ingest.New(client, testConfig.Database, tableName)
+	if err != nil {
+		panic(err)
+	}
+
+	// Create an empty reader
+	reader, writer := io.Pipe()
+	writer.Close()
+
+	// Once without status reporting
```
element-of-surprise, line 309: Make table driven

element-of-surprise (changes requested):

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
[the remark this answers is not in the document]
yogilad: woops.
Removed and added to gitignore

kusto/ingest/internal/filesystem/filesystem.go
```diff
@@ -107,24 +107,25 @@ func (i *Ingestion) Local(ctx context.Context, from string, props properties.All
 }
 
 // Reader uploads a file via an io.Reader.
-func (i *Ingestion) Reader(ctx context.Context, reader io.Reader, props properties.All) error {
+// If the function succeeds, it returns the path of the created blob.
+func (i *Ingestion) Reader(ctx context.Context, reader io.Reader, props properties.All) (string, error) {
```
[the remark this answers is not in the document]
yogilad: It gets added to the status report sent to the cluster.
We can either return the path or pass the result object to be updated

kusto/ingest/internal/properties/properties.go
```diff
@@ -120,6 +120,32 @@ var dfDescriptions = []dfDescriptor{
...
 }
 
+// IngestionReportLevel defines which ingestion statuses are reported by the DM.
+type IngestionReportLevel int
+
+const (
+	// FailuresOnly tells to the DM to report the ingestion sytatus of failed ingestions only.
+	FailuresOnly IngestionReportLevel = 0
+	// None tells to the DM not to report ingestion status.
+	None IngestionReportLevel = 1
```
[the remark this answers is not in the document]
yogilad: True and this pisses me off as well, but we are locked to the current implementation 👍

kusto/ingest/result.go
yogilad: I did this so that I can return res.PutProps() in a single line, but this is irrelevant now

kusto/ingest/result.go
```diff
@@ -0,0 +1,172 @@
...
+		return r
+	}
+
+	r.tableClient = client
+	return r
+}
+
+// Wait returns a channel that can be checked for ingestion results.
+// In order to check actual status please use the IngestionStatus option when ingesting data.
+func (r *Result) Wait(ctx context.Context) chan error {
```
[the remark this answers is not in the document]
yogilad: Dude!
It was like this before, and you requested it to be inside the go routine and made a passionate argument about it costing nothing and being more clear.
:/

kusto/ingest/result.go
yogilad: Again, this is how this was before :/ 
Poll got the channel and managed it.

Anyway, I prefer to keep this as is since there's a good separation of concerns here.
Poll gets the status 
Wait decides what to do with it

kusto/test/etoe/ingestion_status_e2e_test.go
yogilad: I disagree with this test philosophy.
Tests should check one case and report it.
Table driven fails on first test and does not uncover further issues.

kusto/test/etoe/ingestion_status_e2e_test.go
yogilad: See above

kusto/test/etoe/ingestion_status_e2e_test.go
yogilad: I don't see what's the big difference with the exception of calling make.
Also, your example has an error since wait returns a channel not an error, so I still need the channel array or I loose parallelism.

kusto/test/etoe/ingestion_status_e2e_test.go
yogilad: For instance it's very hard for me right now to debug a specific case that does not ingest well in the old e2e code.
I have to let the loop run it's course and check which case is the relevant one.

kusto/test/etoe/etoe_test.go
```diff
@@ -1,15 +1,11 @@
-package main
+package etoe
```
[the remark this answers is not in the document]
element-of-surprise: Out of curiosity, what was your compilation issue

kusto/test/etoe/ingestion_status_e2e_test.go
element-of-surprise: Fixed the bug, I think.  Same code, I just changed it to use a chan error.
The difference is: less loops, doesn't use arrays (which is uncommon) and a single slice instead of two arrays.

But not worth spending a bunch of time on, up to you.

kusto/test/etoe/ingestion_status_e2e_test.go
element-of-surprise: The etoe tests are not a really great version of Go testing.  It was a rough conversion of some test code that was written in a different coding style.  It roughly fits in here.  So there are ugly edges here.

Table driven tests are the standard Go method of testing.  Where it makes sense (testing more than one option), we should default to how everyone in Go does things. 

Table driven tests only fail on the first test if you choose for it to do so.  A table driven test can tests all cases and report only on ones that fail.  Not to mention there is a whole way of doing subtests (though I don't use this facility much, probably because it didn't exist when I started).  

However, at this point, I think the etoe needs the next stage of love to go from the, "translate from .Net testing to rough Go testing", and I can just do this in a later CL.

kusto/ingest/result.go
element-of-surprise: I'm bouncing around with this, why we diff on this.  I get your point of view, at least on the extra "error".  But it seems unclean, and I know I haven't seen that.

But maybe my memory is faulty.  I wanted prior art before I commented back.  So I went digging in the standard library, but wasn't finding anything (they don't really deal a lot with status codes via errors, they handle it with explicit errors).  

I think maybe I got this from stubby, so went and looked at gRPC status calls and found: 
https://github.com/grpc/grpc-go/blob/master/status/status.go#L100

This code base is doing what we are doing (except we split status and failure instead of just a single code).  They return "error" that can later be converted to its *Status type if wanted (we don't do this, yet) or have its Code retrieved.  

So, let's sit and chat about what we should do here.  I'd like to talk the pro's and cons, as I don't feel great about either one.

kusto/ingest/result.go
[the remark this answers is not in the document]
element-of-surprise: Goroutines do cost almost nothing.  I remember that, but I think I was optimizing more than that (though honestly, it could be that my brain is just "farting" between what it likes on different weeks, or lack of coffee, or that i'm just an unreasonable cuss that likes to nitpick).

kusto/ingest/status.go
element-of-surprise: Gotcha, thanks for clarifying.

Issue #35, opened 2020-09-11
Timespan value string marshal is providing nanosecond accuracy, but should only provide "tick" accuracy


element-of-surprise:
The Timespan type is a Go wrapper around time.Duration, which must convert to the string format of .Net's Timespan's type.  Like any other .Net thing, its overly complicated and supports multiple string layouts.  Also, like other .Net types, it defines it's own base type called a "tick" that no one else uses.  A tick is 10 millionth of a second.  Because of these special types and layouts, time.Time.Format() won't work.

Because there are multiple Timespan output formats, I'm not sure if I based mine on a Timespan format that supported nanosecond, or the parsers were more forgiving or what.  The backend was honoring it, probably by chopping off the excess integers.  But some change to the backend started making it not accept the values if the precision was more than a tick.

We need to change the marshal to remove precision of more than a tick.   We can keep Unmarshal's nanosecond accuracy, as it will have no effect on us.

PR #36, opened 2020-09-11
Updates Timespan for Issue #35

element-of-surprise:
#### Pull Request Description

- Makes Timespan.Marshal() use "tick" accuracy instead of Nanosecond.
- Fixes a padding bug due to a copy/paste error

#### Future Release Comment
This can added as a minor release in the current release branch.

**Breaking Changes:**
If a user was using the Timespan.Marshal() for value compares.  This is unlikely.

**Features:**
- None

**Fixes:**
A bug in timespan that the backend is no longer ignoring.

### DIFF
--- a/kusto/data/value/value.go
+++ b/kusto/data/value/value.go
@@ -467,6 +467,8 @@ func (d *Decimal) Unmarshal(i interface{}) error {
 	return nil
 }
 
+const tick = 100 * time.Nanosecond
+
 // Timespan represents a Kusto timespan type.  Timespan implements Kusto.
 type Timespan struct {
 	// Value holds the value of the type.
@@ -534,12 +536,15 @@ func (t Timespan) Marshal() string {
 	seconds := val / time.Second
 	val = val - (seconds * time.Second)
 	switch {
-	case minutes < 10:
+	case seconds < 10:
 		sb.WriteString(fmt.Sprintf("0%d", int(seconds)))
 	default:
 		sb.WriteString(fmt.Sprintf("%d", int(seconds)))
 	}
 
+	// In this section we build a subSecond string representation.
+	// This is compbined later with the existing content if and only if
+	// we have a subSecond value (which is determined by set==true).
 	subSecond := strings.Builder{}
 	set := false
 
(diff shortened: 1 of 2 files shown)


kusto/data/value/value.go
```diff
@@ -534,12 +536,15 @@ func (t Timespan) Marshal() string {
...
 	switch {
-	case minutes < 10:
+	case seconds < 10:
 		sb.WriteString(fmt.Sprintf("0%d", int(seconds)))
 	default:
 		sb.WriteString(fmt.Sprintf("%d", int(seconds)))
 	}
 
+	// In this section we build a subSecond string representation.
+	// This is compbined later with the existing content if and only if
```
yogilad, line 546: type

kusto/data/value/value.go
```diff
@@ -534,12 +536,15 @@ func (t Timespan) Marshal() string {
 	seconds := val / time.Second
 	val = val - (seconds * time.Second)
 	switch {
-	case minutes < 10:
+	case seconds < 10:
 		sb.WriteString(fmt.Sprintf("0%d", int(seconds)))
```
yogilad, line 540: You can use %02d which will pad the printed value with zeroes.
This can simply most of this function and especially the fractional part

kusto/data/value/value.go
element-of-surprise: Comment removed with changes.

kusto/data/value/value.go
element-of-surprise: Change and code simplified.  Completely forgot about this, good catch.

yogilad (changes requested):

kusto/data/value/value.go
```diff
@@ -497,82 +497,44 @@ func (t Timespan) Marshal() string {
...
+	// Add a - sign if we have a negative value. Convert our value to positive for easier processing.
 	if t.Value < 0 {
 		sb.WriteString("-")
 		val = val * -1
 	}
 
+	// Only include the day if the duration is 1+ days.
 	days := val / day
 	val = val - (days * day)
 	switch {
```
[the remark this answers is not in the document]
element-of-surprise: There are several documents on this time format and several different versions of the time format output.  

I am conforming to the "c" format, listed as:
[-][d'.']hh':'mm':'ss['.'fffffff]

Ref: https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-timespan-format-strings

There is a "g" format that is a format more like what you are talking about, but it requires I support "culture sensitive" information.  I do not want to do that.

As the days are an optional value as well as the subseconds, we are supposed to remove those values if they are zero, hence the switch statement and the trailing 0s code.

I am padding the days, which is not needed, changed.

kusto/data/value/value.go
```diff
@@ -495,79 +498,45 @@ func (t Timespan) Marshal() string {
...
-	if set {
-		sb.WriteString("." + subSecond.String())
+	ticks := val / tick
+	if milliseconds > 0 || ticks > 0 {
+		sb.WriteString(fmt.Sprintf(".%03d%d", milliseconds, ticks))
 	}
 
+	// Remove any trailing 0's.
 	str := strings.TrimRight(sb.String(), "0")
 	if strings.HasSuffix(str, ":") {
```
yogilad, line 541: This is an optional comment.
You can trim the fractional part inside the if then attach it, instead of this "oops, let's fix this" condition.

yogilad (approved):

PR #45, opened 2021-03-15
Add Managed functions

AsafMah:
Added functions that act like StreamingManagedIngestion.

It's still WIP (needs tests), but would like to know what do you think about the design

### DIFF
--- a/kusto/data/errors/errors.go
+++ b/kusto/data/errors/errors.go
@@ -103,7 +103,7 @@ func (e *Error) UnmarshalREST() map[string]interface{} {
 	return m
 }
 
-// SetNoRetry sets this error so that ShouldRetry() will always return false.
+// SetNoRetry sets this error so that Retry() will always return false.
 func (e *Error) SetNoRetry() *Error {
 	e.permanent = true
 	return e
@@ -165,9 +165,9 @@ func (e *Error) Error() string {
 	return b.String()
 }
 
-// ShouldRetry determines if the error is transient and the action can be retried or not.
+// Retry determines if the error is transient and the action can be retried or not.
 // Some errors that can be retried, such as a timeout, may never succeed, so avoid infinite retries.
-func ShouldRetry(err error) bool {
+func Retry(err error) bool {
 	var e *Error
 	if errors.As(err, &e) {
 		// e.permanent can be set multiple ways. If it is true, you can never retry.
@@ -189,16 +189,11 @@ func ShouldRetry(err error) bool {
 		}
 
 		if e.inner != nil {
-			return ShouldRetry(e.inner)
+			return Retry(e.inner)
 		}
 		return true
 	}
 	return false
-}
-
-// Deprecated: replaced by ShouldRetry, which is a clearer name
-func Retry(err error) bool {
-	return ShouldRetry(err)
 }
 
 // E constructs an Error. You may pass in an Op, Kind and error.  This will strip a *errors.Error(the error in this package) if you
--- a/kusto/data/errors/errors.go
+++ b/kusto/data/errors/errors.go
@@ -103,7 +103,7 @@ func (e *Error) UnmarshalREST() map[string]interface{} {
 	return m
 }
 
-// SetNoRetry sets this error so that Retry() will always return false.
+// SetNoRetry sets this error so that ShouldRetry() will always return false.
 func (e *Error) SetNoRetry() *Error {
 	e.permanent = true
 	return e
@@ -165,9 +165,9 @@ func (e *Error) Error() string {
 	return b.String()
 }
 
-// Retry determines if the error is transient and the action can be retried or not.
+// ShouldRetry determines if the error is transient and the action can be retried or not.
 // Some errors that can be retried, such as a timeout, may never succeed, so avoid infinite retries.
-func Retry(err error) bool {
+func ShouldRetry(err error) bool {
 	var e *Error
 	if errors.As(err, &e) {
 		// e.permanent can be set multiple ways. If it is true, you can never retry.
@@ -189,11 +189,16 @@ func Retry(err error) bool {
 		}
 
 		if e.inner != nil {
-			return Retry(e.inner)
+			return ShouldRetry(e.inner)
 		}
 		return true
 	}
 	return false
+}
+
+// Deprecated: replaced by ShouldRetry, which is a clearer name
+func Retry(err error) bool {
+	return ShouldRetry(err)
 }
 
 // E constructs an Error. You may pass in an Op, Kind and error.  This will strip a *errors.Error(the error in this package) if you
--- a/kusto/ingest/ingest.go
+++ b/kusto/ingest/ingest.go
@@ -1,6 +1,7 @@
 package ingest
 
 import (
+	"bufio"
 	"bytes"
 	"compress/gzip"
 	"context"
@@ -14,7 +15,6 @@ import (
 	"github.com/Azure/azure-kusto-go/kusto/ingest/internal/resources"
 	"github.com/google/uuid"
 	"io"
-	"io/ioutil"
 	"os"
 	"sync"
 	"sync/atomic"
@@ -43,10 +43,6 @@ func getManager(client *kusto.Client) (*resources.Manager, error) {
 		return mgr, nil
 	}
 	return i.(*resources.Manager), nil
-}
-
-type streamer interface {
-	stream(ctx context.Context, format DataFormat, mappingName string) (io.WriteCloser, error)
 }
 
 // Ingestion provides data ingestion from external sources into Kusto.
@@ -331,11 +327,6 @@ func FileFormat(et DataFormat) FileOption {
 	)
 }
 
-type mapEntry struct {
-	Name string
-	Kind string
-}
-
 func (i *Ingestion) prepForIngestion(ctx context.Context, options []FileOption) (*Result, properties.All, error) {
 	result := newResult()
 
@@ -442,7 +433,7 @@ var (
 )
 
 const (
-	MaxRetryCount    = 3
+	maxRetryCount    = 3
 	mib              = 1024 * 1024
 	MaxStreamingSize = 4 * mib
 )
@@ -451,7 +442,7 @@ const (
 // This method is thread-safe.
 func (i *Ingestion) FromFileManaged(ctx context.Context, fPath string,
 	options ...FileOption) (*Result, error) {
-	result, props, err := i.prepForIngestion(ctx, options)
+	result, _, err := i.prepForIngestion(ctx, options)
 	if err != nil {
 		return nil, err
 	}
@@ -463,7 +454,7 @@ func (i *Ingestion) FromFileManaged(ctx context.Context, fPath string,
 		return nil, err
 	}
 
-	skipStreaming := false
+	skipStreaming := true
 
 	if local {
 		stat, err := os.Stat(fPath)
@@ -475,16 +466,7 @@ func (i *Ingestion) FromFileManaged(ctx context.Context, fPath string,
 
 	if skipStreaming { // In case of blob it's inefficient to use streaming ingest,
 		// so we're defaulting to queued
-		if local {
-			err = i.fs.Local(ctx, fPath, props)
-		} else {
-			err = i.fs.Blob(ctx, fPath, 0, props)
-		}
-		if err != nil {
-			return nil, err
-		}
-		result.putQueued(i.mgr)
-		return result, nil
+		return i.FromFile(ctx, fPath, options...)
 	}
 
 	file, err := os.Open(fPath)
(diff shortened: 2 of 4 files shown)


kusto/ingest/ingest.go
```diff
@@ -440,7 +441,87 @@ var (
...
+	result.record.IngestionSourcePath = fPath
+
+	local, err := filesystem.IsLocalPath(fPath)
+	if err != nil {
+		return nil, err
+	}
+
+	skipStreaming := false
+
+	if local {
```
yogilad, line 468: And if it is a huge blob?
Not sure if it can be provided, but do you check the ingestion properties for the file size?
If there's no such option in Go maybe we need to support it or at least request this value in the Managed APIs.

kusto/ingest/ingest.go
```diff
@@ -440,7 +441,87 @@ var (
...
+		stat, err := os.Stat(fPath)
+		if err != nil {
+			return nil, err
+		}
+		skipStreaming = stat.Size() > MaxStreamingSize
+	}
+
+	if skipStreaming { // In case of blob it's inefficient to use streaming ingest,
```
yogilad, line 476: Call FromFile()

kusto/ingest/ingest.go
```diff
@@ -440,7 +441,87 @@ var (
...
+			err = i.fs.Blob(ctx, fPath, 0, props)
+		}
+		if err != nil {
+			return nil, err
+		}
+		result.putQueued(i.mgr)
+		return result, nil
+	}
+
+	file, err := os.Open(fPath)
```
yogilad, line 490: What happens if this is an Azure blob?

kusto/ingest/ingest.go
```diff
@@ -440,7 +441,87 @@ var (
...
+}
+
+func (i *Ingestion) FromReaderManaged(ctx context.Context, reader io.Reader,
+	options ...FileOption) (*Result, error) {
+	result, props, err := i.prepForIngestion(ctx, options)
+	if err != nil {
+		return nil, err
+	}
+
+	bytesRead, err := ioutil.ReadAll(reader)
```
yogilad, line 505: The unknown size of blobs escalates here

kusto/ingest/ingest.go
```diff
@@ -440,7 +441,87 @@ var (
...
+		return nil, err
+	}
+
+	for j := 0; j < MaxRetryCount; j++ {
+		err = i.Stream(ctx, bytesRead, props.Ingestion.Additional.Format, props.Ingestion.Additional.IngestionMappingRef)
+		if err == nil {
+			result.record.Status = Succeeded
+			return result, nil
+		}
+		if errors.Retry(err) {
```
yogilad, line 516: ShouldRetry or Retryable

kusto/ingest/ingest.go
```diff
@@ -440,7 +441,87 @@ var (
...
+			time.Sleep(time.Duration(j) * time.Second)
```
yogilad, line 517: If the error is Throttle, you may want to wait a bit longer

kusto/ingest/ingest.go
```diff
@@ -440,7 +441,87 @@ var (
...
-const mib = 1024 * 1024
+const (
+	MaxRetryCount    = 3
+	mib              = 1024 * 1024
+	MaxStreamingSize = 4 * mib
+)
+
+// FromFile allows uploading a data file for Kusto from either a local path or a blobstore URI path.
+// This method is thread-safe.
+func (i *Ingestion) FromFileManaged(ctx context.Context, fPath string,
```
yogilad, line 452: Is the user aware in other SDKs if the data was streamed or queued?
You may want to support the result model of FromFile/Reader if you need to allow queued ingestion status checking.

kusto/ingest/ingest.go
```diff
@@ -466,7 +547,7 @@ func (i *Ingestion) Stream(ctx context.Context, payload []byte, format DataForma
 	if err := zw.Close(); err != nil {
 		return errors.E(errors.OpIngestStream, errors.KClientArgs, err).SetNoRetry()
 	}
-	if buf.Len() > 4*mib {
+	if buf.Len() > MaxStreamingSize {
```
yogilad, line 550: Mind this is a breaking change

kusto/ingest/ingest.go
```diff
@@ -440,7 +441,87 @@ var (
 	ErrTooLarge = errors.ES(errors.OpIngestStream, errors.KClientArgs, "cannot add data larger than 4MiB")
 )
 
-const mib = 1024 * 1024
+const (
+	MaxRetryCount    = 3
+	mib              = 1024 * 1024
```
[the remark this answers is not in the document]
AsafMah: The mib was here before me, I assume that it's to clarify that it's Mebibyte, but I agree Mb would work just as well

kusto/ingest/ingest.go
AsafMah: If it's a blob it's automatically queued Ingestion, because downloading it only to send is inefficient.

If it's a local file we check it's size via Stat

kusto/ingest/ingest.go
AsafMah: That re-does many of the things this function does, so it's a bit inefficient

kusto/ingest/ingest.go
AsafMah: Then it will do queued Ingestion
Skipstreaming is always true for blobs

kusto/ingest/ingest.go
AsafMah: The problem is that the interface doesn't let us to get the size before hand.
I'll look into properties or other methods

kusto/ingest/ingest.go
AsafMah: This methods existed before this pr, changing its name is technically a breaking change

kusto/ingest/ingest.go
AsafMah: How? MaxStreamingSize is defined to be 4*mib

kusto/ingest/ingest.go
yogilad: Where in the code is it redirected to queued?

kusto/ingest/ingest.go
yogilad: Better to reuse in most cases.
is there anything of significance that gets redone?
Also, you may need it for the result status reporting

kusto/ingest/ingest.go
yogilad: I think your default for skipStreaming is incorrect in that case

kusto/ingest/ingest.go
yogilad: You change the behavior of an existing API.
Someone streaming 5MB today would fail

kusto/ingest/ingest.go
AsafMah: In the code inside if skipStreaming

kusto/ingest/ingest.go
AsafMah: The if was already there, all this change is giving the constant a name.

yogilad: LOL :D 
I completely ignored this somehow

kusto/ingest/ingest.go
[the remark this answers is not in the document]
element-of-surprise: It is to clarify mebibyte.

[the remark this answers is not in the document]
element-of-surprise: Also, why is MaxRetryCount public?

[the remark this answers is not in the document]
AsafMah: oh yeah, I forgot about go's access rules

kusto/ingest/ingest.go
AsafMah: I added a WasStreamed property to the result object that you can check against

kusto/ingest/ingest.go
AsafMah: Now the skip streaming is correct

kusto/ingest/ingest.go
AsafMah: Fair enough, now it's done

kusto/ingest/ingest.go
AsafMah: Fixed

kusto/ingest/ingest.go
AsafMah: I now changed it to ShouldRetry and kept Retry as a deprecated alias

kusto/ingest/ingest.go
AsafMah: Now I use the props, and peek to never actually read more than the max size. What do you think?

kusto/ingest/ingest.go
AsafMah: No other ManagedIngest does that, can you point me to what a throttle error would look like on the client side so I can implement this?

kusto/data/errors/errors.go
```diff
@@ -165,9 +165,9 @@ func (e *Error) Error() string {
 	return b.String()
 }
 
-// Retry determines if the error is transient and the action can be retried or not.
+// ShouldRetry determines if the error is transient and the action can be retried or not.
 // Some errors that can be retried, such as a timeout, may never succeed, so avoid infinite retries.
-func Retry(err error) bool {
+func ShouldRetry(err error) bool {
```
element-of-surprise, line 170: While I get your reasoning, the case for making the change is disrupted by the additional function calls that are now add confusion makes the API worse than the cure.

Please revert the name change.

kusto/ingest/ingest.go
```diff
@@ -1,22 +1,24 @@
...
 	"github.com/Azure/azure-kusto-go/kusto/ingest/internal/conn"
 	"github.com/Azure/azure-kusto-go/kusto/ingest/internal/filesystem"
 	"github.com/Azure/azure-kusto-go/kusto/ingest/internal/properties"
 	"github.com/Azure/azure-kusto-go/kusto/ingest/internal/resources"
 	"github.com/google/uuid"
+	"io"
+	"os"
+	"sync"
+	"sync/atomic"
+	"time"
```
element-of-surprise, line 21: Please put these back above. I prefer a:

```
stdlib
...

internal libs
...

third_party
...

renames
...
```

kusto/ingest/ingest.go
```diff
@@ -1,22 +1,24 @@
 package ingest
```
element-of-surprise, line 1: I had this conversation yesterday with Yochai in that this managed thing looks like something that should just be rolled into the existing structure or replace it.  This managed thing also seems to be the frontend trying to fix a backend problem.   I guess it is what it is.....

I'm not as interested in how this looks in the other APIs as much as making sure the user can use the feature.  And I'm super interested in a clean API with the least amount of junk.  We should avoid unnecessary API breakage, but we are a 0.X release, so we can do so if the benefit is high enough.

@AsafMah : I'd like to think about how to roll as many of these methods together as possible.  Unless we MUST keep them separate because of functional reasons, I'd like to combine these and not multiply them.  We can use options to specify specific settings (like Managed()). 

My ask is:
@AsafMah : forgetting the code implementation, can you propose an API with the least moving parts that supports our current cases + managed?

AsafMah: My big problem with making this an option is discoverability, most people won't know that managed ingestion is a thing and probably won't use it.

Other SDKs have three separate client types for different types of ingestion - Queued, Streaming and ManagedStreaming
Since we only have one client in go that supports both streaming and queued, I thought that adding managed functions is a good middleground.

The options as I see them are -
1. Split it to a different client struct - we don't want that
2. Have "managed" be an additional parameter to the functions we have - go doesn't support default parameters or overloading, so this breaks the interface
3. Add managed as a FileOption - Discoverability + it's not really a "file option"
4. Add it as an option and make it the default - we probably don't want to change the default behaviour
4. Add new Managed functions - I agree it's not ideal, but I think it's best considering the circumstances

yogilad: @element-of-surprise , since you already plan on breaking and remaking the SDK, lets have two APIs calling one hidden impl.
When you've rearchitected the SDK, feel free to drop the old streaming API.
WDYT?

kusto/ingest/ingest.go
element-of-surprise: I think I put this somewhere else, bu just in case:  let's not do this.

kusto/ingest/ingest.go
element-of-surprise: @AsafMah 

#3 seems more ideal to me, and I don't agree with the premises in that discoverability is a problem.  A Go user is either going to look at the options if they are interested or search for "Managed" in the docs if they know its there.

#4 you can do default values in Go, you just don't do it with "attribute=Default" in the definition of some method.

FileOption was chosen as the name when we just had Stream and FromFile().  It was notating that it just changed files, not that it matters all that much. The user just looks at the type and then looks for what implements the options (in Go).  I honestly don't even see the type names for options.

My mind is starting to coalesce around a single main method for doing everything, with options.  That method would be:

```go
 FromReader(ctx context.Context, reader io.Reader, options ...ReaderOption) (*Result, error)
```

FromReader, given the right arguments, can do everything.  This could handle input from a stream, memory, or a file.  Given the right options.  To me, Managed() just becomes an option at that point.  

We could then add our individual FromFile(), FromStream() directly on top of that or we could fork out a "high level" package that had these.  They would have their own options that would put in ReaderOption(s).  Some of those options might be "hidden"  for FromReader buy exposed by our high levels.

@yogilad 

So it depends on how disruptive you want to be.  What you are talking about doing is disrupting in a few different ways:

If we make changes now:
FromReader() (or maybe just calling it From()) having some minor changes.
FromFile() having minor changes, possibly a location move
FromStream() having minor changes, possibly a location move

If we do it later:
Removal of Managed* + those items above.  
User must figure out that Managed* became an option and those methods are gone.

@yogilad and @AsafMah 

We probably need to sit down and discuss this.  We are coming at this from two different directions with the same goal.  I'm super hard on APIs, because changes are usually either disruptive or add baggage.  

If managed is a thing our clients need, then I want it in the API.

But I don't want to start exploding the API surface to get there.  I heard something once that resinates with me: " if you know something well, you don't add features, you take them away.  "  I'm paraphrasing, but I think it goes with "Less is More" from Pike:  https://commandcenter.blogspot.com/2012/06/less-is-exponentially-more.html

kusto/ingest/ingest.go
AsafMah: don't worry it's already been reverted :)

kusto/ingest/ingest.go
yogilad: If that's the case then let's forget this

PR #47, opened 2021-05-03
Cache kusto identity tokens for 1h

yogilad:
#### Pull Request Description

Cache kusto identity tokens for 1h

---

#### Future Release Comment
**Fixes:**
- Ingestion does not call the cluster for identity token for each file/reader

### DIFF
--- a/kusto/ingest/internal/resources/resources.go
+++ b/kusto/ingest/internal/resources/resources.go
@@ -126,9 +126,11 @@ type token struct {
 
 // Manager manages Kusto resources.
 type Manager struct {
-	client    mgmter
-	done      chan struct{}
-	resources atomic.Value // Stores Ingestion
+	client               mgmter
+	done                 chan struct{}
+	resources            atomic.Value // Stores Ingestion
+	kustoToken           token
+	kustoTokenExpiration time.Time
 }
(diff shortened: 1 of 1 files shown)


AsafMah (approved):
Looks good

But it probably does need a test, and I'm not sure how it can be done with 1 hour hardcoded

yogilad:
@AsafMah , I manually tested this with a 10s expiration period to see that we go through > not cached => cached => expired => cached

kusto/ingest/internal/resources/resources.go
```diff
@@ -126,9 +126,11 @@ type token struct {
...
 // Manager manages Kusto resources.
 type Manager struct {
-	client    mgmter
-	done      chan struct{}
-	resources atomic.Value // Stores Ingestion
+	client               mgmter
+	done                 chan struct{}
+	resources            atomic.Value // Stores Ingestion
+	kustoToken           token
+	kustoTokenExpiration time.Time
```
yihezkel, line 133: I know this PR is already merged, but small comment -
"kustoTokenExpiration" sounds like it's when the token expires. I suggest "kustoTokenCacheExpiration".

PR #52, opened 2021-07-01
Azure Synapse Audience Support

yogilad:
#### Pull Request Description

Use fixed audience for URIs having azuresynapse like segments in their URIs

---

#### Future Release Comment

**Features:**
- Support authenticating to Synapse Kusto clusters

### DIFF
--- a/kusto/kusto.go
+++ b/kusto/kusto.go
@@ -130,14 +130,13 @@ func New(endpoint string, auth Authorization, options ...Option) (*Client, error
 		o(client)
 	}
 
+	validationEndpoint := endpoint
 	if strings.Contains(strings.ToLower(endpoint), ".azuresynapse") {
-		if err := auth.Validate("https://kusto.kusto.windows.net"); err != nil {
-			return nil, err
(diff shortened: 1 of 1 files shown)


kusto/kusto.go
```diff
@@ -130,9 +130,16 @@ func New(endpoint string, auth Authorization, options ...Option) (*Client, error
 		o(client)
 	}
 
-	if err := auth.Validate(endpoint); err != nil {
-		return nil, err
+	if strings.Contains(strings.ToLower(endpoint), ".azuresynapse") {
```
AsafMah, line 133: WDYT about writing the like this:

```go 
validateEndpoint := endpoint
if strings.Contains(strings.ToLower(endpoint), ".azuresynapse") {
    validateEndpoint = "https://kusto.kusto.windows.net"
}
if err := auth.Validate(validateEndpoint); err != nil {
		return nil, err
}
```

To save on nesting and calling the method twice

AsafMah (approved):

PR #60, opened 2021-11-25
Updated azure-storage-blob-go to be able to use a TransferManager that requires less memory

AsafMah:
#### Pull Request Description

This comes from reports on telegraf that we take hundreds of megs of RAM.
Turns out, this is true.

We have "magic numbers" in the code for optimal work with azure blobs.
The numbers are 8mb * 50 concurrency = 400mb that were allocated when uploading blobs.

Luckily this issue was addressed in newer versions of azblob (thanks to @element-of-surprise) - https://github.com/Azure/azure-storage-blob-go/pull/234

So we now use a SyncPool TransferManager, which reduces our RAM 10x times.

It's still more than other telegraf plugins, but for that we will need to change the "magic numbers", to which John wrote -
```
	// The numbers below are magic numbers. They were derived from doing Azure to Azure tests of azcopy for various file sizes
	// to prove that changes weren't going to make azcopy slower. It was found that multipying azcopy's concurrency by 10x (to 50)
	// made a 5x improvement in speed. We don't have any numbers from the service side to give us numbers we should use, so this
	// is our best guess from observation. DO NOT CHANGE UNLESS YOU KNOW BETTER.

	blockSize   = 8 * _1MiB
	concurrency = 50
```

---

#### Future Release Comment
Fixed high ram usage from azure blob upload

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
-Fixed high ram usage from azure blob upload


kusto/ingest/internal/filesystem/filesystem.go
```diff
@@ -53,16 +53,28 @@ type Ingestion struct {
 
 	uploadBlobStream uploadBlobStream
 	uploadBlobFile   uploadBlobFile
+	transferManager  azblob.TransferManager
 }
 
 // New is the constructor for Ingestion.
 func New(db, table string, mgr *resources.Manager) (*Ingestion, error) {
```
yogilad, line 60: Is anyone using the old API?
If not consider merging

kusto/ingest/ingest.go
```diff
@@ -64,12 +64,17 @@ type Ingestion struct {
 
 // New is the constructor for Ingestion.
 func New(client *kusto.Client, db, table string) (*Ingestion, error) {
+	return NewWithBlobProperties(client, db, table, filesystem.BlockSize, filesystem.Concurrency)
+}
+
+// NewWithBlobProperties is an advanced constructor for Ingestion, to use in special cases where you need more control over memory consumption.
```
yogilad, line 70: Not a fan of the name - it does not really explain what this new command does.
Nor the parameters which need to be explained in the comment.
Do you have the information or can you dig up a better explanation for the user?

yogilad (approved):

kusto/ingest/ingest.go
```diff
@@ -64,12 +64,17 @@ type Ingestion struct {
 
 // New is the constructor for Ingestion.
 func New(client *kusto.Client, db, table string) (*Ingestion, error) {
+	return NewWithBlobProperties(client, db, table, filesystem.BlockSize, filesystem.Concurrency)
```
element-of-surprise, line 67: We shouldn't do this or expose a NewWithBlobProperties type.  

We should simply add options.

```go
func New(client *kusto.Client, db, table string, options ...Option) (*Ingestion, error) {
  i := &Ingestion{...}
  for _, o := range options  {
    o(i)
  }
  ...
}
```
Please reach out to me if this isn't clear.

kusto/ingest/ingest.go
element-of-surprise: Let's remove this completely, there is no reason to have a new constructor.

kusto/ingest/internal/filesystem/filesystem.go
```diff
@@ -34,8 +34,8 @@ const (
 	// made a 5x improvement in speed. We don't have any numbers from the service side to give us numbers we should use, so this
 	// is our best guess from observation. DO NOT CHANGE UNLESS YOU KNOW BETTER.
 
-	blockSize   = 8 * _1MiB
-	concurrency = 50
+	BlockSize   = 8 * _1MiB
+	Concurrency = 50
```
element-of-surprise, line 38: These should stay private.  We can document in the constructor options what the default values are.

element-of-surprise (commented):
As a general note, we should probably consider removing our vendor directory.  This is the older style of doing this and we should consider just controlling this with go.mod at this point.  

I think that should probably be done in its own PR.

kusto/ingest/internal/filesystem/filesystem.go
AsafMah: Irrelevant now since we added options

kusto/ingest/ingest.go
AsafMah: I now added the options pattern.
It got a bit messy because I needed to add it to both `filesystem` and `ingest`,
and also because NewSyncPool needs the arguments at the ctor.

What do you think? (also @yogilad )

kusto/ingest/internal/filesystem/filesystem.go
```diff
@@ -53,17 +53,57 @@ type Ingestion struct {
...
 		uploadBlobFile:   azblob.UploadFileToBlockBlob,
 	}
+
+	for _, opt := range options {
+		opt(i)
+	}
+
+	if i.bufferSize == 0 {
+		i.bufferSize = blockSize
+	}
```
yogilad, line 95: else default?

kusto/ingest/internal/filesystem/filesystem.go
```diff
@@ -53,17 +53,57 @@ type Ingestion struct {
...
+
+	if i.concurrency == 0 {
+		i.concurrency = concurrency
+	}
```
yogilad, line 99: same

yogilad (approved):

kusto/ingest/ingest.go
```diff
@@ -60,16 +60,31 @@ type Ingestion struct {
...
+		s.blockSize = size
 	}
+}
+
+// WithConcurrency sets the concurrency for uploading streams to kusto. Defaults to 50.
+func WithConcurrency(concurrency int) Option {
+	return func(s *Ingestion) {
+		s.concurrency = concurrency
+	}
+}
```
element-of-surprise, line 84: This is better, so thanks for the changes.

My thinking is that we should not hide the TransferManager.  By just providing two knobs, the user can only turn the two knobs inside a single TransferManager.  We can give them much more control by simply letting them specify the TransferManager they want to use and telling them what the default is.  If they are really concerned about memory, they can simply use the static buffer TM.  They can even create their own that satisfies their needs.

```go
// WithTransferManager provides a custom azblob TransferManager for copying our ingest data to
// Azure Blob Storage for ingestion into Kusto during non-streamed ingestion.  By default 
// we use an azblob.SyncPool that uses 8MiB blocks and 50 concurrent connections.   You can create a
// one with your own settings using azblob.NewSyncPool() or the static allocation with azblob. NewStaticBuffer().
func WithTransferManager(t *azblob.TransferManager) Option {
  return func(s * Ingestion) {
    s.tm = t
  }
}
```

Now in New(), we would do:
```go

func New(client *kusto.Client, db, table string, options ...Option) (*Ingestion, error) {
        mgr, err := getManager(client)
	if err != nil {
		return nil, err
	}

        i := &Ingestion{
		client: client,
		mgr:    mgr,
		db:     db,
		table:  table,
		fs:     fs,
	}

        for _, option := range options {
		option(i)
	}

        if i.tm == nil {
            tm, err := azblob.NewSyncPool(blockSize, concurrency)
	    if err != nil {
		return nil, err
	    }
       }
       ....
```
You would then change filesystem.New() to simply take a TransferManager.  It doesn't need options at this point, as its internal/, so no user can import it.  This code also assumes we move blockSize and concurrency to ingest.go instead of filesystem.go .

Now the users have the full power of the TransferManager and I think it simplifies this down some.

yogilad: @element-of-surprise , I don't like this change.
It exposes to much of the internal implementation to the user.

The previous commit was far better and needed just a little more touch ups (create structs with file system defaults rather then zero, use the size and concurrency in upload File & stream).
@AsafMah , can you revert and push a change with what we discussed?

yogilad (approved):

kusto/ingest/ingest.go
```diff
@@ -60,27 +60,54 @@ type Ingestion struct {
...
+
+	uploadBlockSize   int
+	uploadConcurrency uint16
+}
+
+// Option is an optional argument to New().
+type Option func(s *Ingestion)
+
+// WithUploadBlockSize sets the block size for uploading blobs to kusto. Defaults to 8MB.
+func WithUploadBlockSize(size int) Option {
```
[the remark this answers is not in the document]
element-of-surprise: As a note, I just made concurrency an int that I would convert, but you can do an uint16.

kusto/ingest/internal/filesystem/filesystem.go
```diff
@@ -295,8 +329,8 @@ func (i *Ingestion) localToBlob(ctx context.Context, from string, to azblob.Cont
 		file,
 		blobURL,
 		azblob.UploadToBlockBlobOptions{
-			BlockSize:   blockSize,
-			Parallelism: concurrency,
+			BlockSize:   int64(i.uploadBlockSize),
+			Parallelism: i.uploadConcurrency,
 		},
 	)
```
element-of-surprise, line 335: As a note, we agreed to remove this in another PR.

kusto/ingest/ingest.go
```diff
@@ -57,24 +57,18 @@ type Ingestion struct {
...
+	bufferSize int
+	maxBuffers int
 }
 
 // Option is an optional argument to New().
 type Option func(s *Ingestion)
 
-// WithUploadBlockSize sets the block size for uploading blobs to kusto. Defaults to 8MB.
-func WithUploadBlockSize(size int) Option {
+// WithStaticBuffer sets a static buffer with a buffer size and max amount of buffers for uploading blobs to kusto.
```
yogilad, line 67:
```suggestion
// WithStaticBuffer configures the ingest client to uploading data to Kusto using a set of static memory buffers with a fixed size up to the defined max amount.
```

kusto/ingest/ingest.go
```diff
@@ -57,24 +57,18 @@ type Ingestion struct {
...
-		s.uploadBlockSize = size
-	}
-}
-
-// WithUploadConcurrency sets the concurrency for uploading blobs to kusto. Defaults to 50.
-func WithUploadConcurrency(concurrency uint16) Option {
-	return func(s *Ingestion) {
-		s.uploadConcurrency = concurrency
+		s.bufferSize = bufferSize
+		s.maxBuffers = maxBuffers
```
yogilad, line 71: Do you want to enforce here or elsewhere a validation check on minimal values for *both* arguments?

kusto/ingest/internal/filesystem/filesystem.go
```diff
@@ -55,44 +55,42 @@ type Ingestion struct {
...
 	}
 
 	for _, opt := range options {
 		opt(i)
 	}
 
-	transferManager, err := azblob.NewSyncPool(i.uploadBlockSize, int(i.uploadConcurrency))
+	var transferManager azblob.TransferManager
+	var err error
+	if i.bufferSize == 0 {
```
yogilad, line 89: Greater then, also check the other value is a non zero positive

yogilad (approved):

kusto/ingest/internal/filesystem/filesystem.go
element-of-surprise: These didn't get marked as private, still are public.

kusto/ingest/internal/filesystem/filesystem.go
```diff
@@ -53,17 +53,51 @@ type Ingestion struct {
...
+
+	uploadBlockSize   int
+	uploadConcurrency uint16
+}
+
+// Option is an optional argument to New().
+type Option func(s *Ingestion)
+
+// WithUploadBlockSize sets the block size for uploading blobs to kusto. Defaults to 8MB.
+func WithUploadBlockSize(size int) Option {
```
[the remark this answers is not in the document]
AsafMah: This (and the next comment) don't exist anymore.
The name is now WithStaticBuffer

kusto/ingest/ingest.go
AsafMah: Now checking both non-zero, as we discussed

kusto/ingest/internal/filesystem/filesystem.go
AsafMah: Added clearer errors, as we discussed

yogilad (approved):

element-of-surprise (approved):

PR #62, opened 2021-12-21
Small fixes

AsafMah:
#### Pull Request Description

This PR fixes some open code quality issues, such as -
Naming in the correct format
Docs with wrong names
Unused parameters
Simplify regexes
don't defer in loop in tests

Shouldn't have any functional changes on the code


kusto/conn.go
```diff
@@ -117,7 +117,7 @@ type execResp struct {
 	frameCh    chan frames.Frame
 }
 
-func (c *conn) execute(ctx context.Context, execType int, db string, query Stmt, payload string, properties requestProperties) (execResp, error) {
```
yogilad, line 120: Mind this is a breaking change despite it not being used.
Please comment on it in the commit.

kusto/ingest/internal/conn/conn_test.go
```diff
@@ -118,58 +124,61 @@ func TestStream(t *testing.T) {
...
+			if err := zw.Close(); err != nil {
+				panic(err)
+			}
+
+			err = conn.Write(ctx, "database", "table", &payload, properties.JSON, test.mappingName)
+
+			switch {
+			case err == nil && test.err:
+				t.Errorf("TestStream(%s): got err == nil, want err != nil", test.desc)
+				return
```
yogilad, line 152: You sure Go supports multiple errors in the same test method?

yogilad (approved):

kusto/conn.go
AsafMah: Is it a breaking change to change a private method?

kusto/ingest/internal/conn/conn_test.go
AsafMah: I don't understand.
It does display multiple errors if there are

yogilad (approved):

PR #69, opened 2022-01-18
Update azure-sdk-for-go version and replace uuid library to fix security bug.

### LINKED ISSUE #51: Vulnerability in github.com/satori/go.uuid
One of the dependencies in this package (github.com/satori/go.uuid) has a vulnerability that dates back quite a while with no published version that has a fix. The repository for that package appears to be abandoned. Here's the CVE: https://snyk.io/vuln/SNYK-GOLANG-GITHUBCOMSATORIGOUUID-72488

Is it possible to upgrade to a different package? I see that there is also a dependency on github.com/google/uuid here. As another reference point, the Azure SDK for Golang chose to move to github.com/gofrs/uuid when faced with the same issue: https://github.com/Azure/azure-sdk-for-go/issues/14283

AsafMah:
Fixes #51


kusto/ingest/status.go
```diff
@@ -5,9 +5,9 @@ import (
 	"time"
 
 	"github.com/Azure/azure-kusto-go/kusto/ingest/internal/properties"
+	storageuid "github.com/gofrs/uuid"
 	"github.com/google/uuid"
 	"github.com/kylelemons/godebug/pretty"
-	storageuid "github.com/satori/go.uuid"
```
yogilad, line 10: I would guess this requires some code changes as IIRC there are explicit references to satori GUID in tableClient

yogilad (approved):

kusto/ingest/status.go
AsafMah: Changed storageuid  to point to the new package

PR #71, opened 2022-02-06
Stop caching manager as it no longer fits the design

AsafMah:
#### Pull Request Description

- Make Ingest accept an interface
- Stop caching manager as it no longer fits the design.

Currently we use a singleton resource manager for every connection. 
This was to save resources of creating one for each connection, but this was when clients didn't have endpoint or authentication.

Now that they do, it causes a bug that all ingestions will be forwarded to the first specified client endpoint and use its authentication.

This PR removes this singleton, and makes a new resource manager for each client, similar to other sdk.

In the future, it may be possible to cache these managers in a correct way, but for now that's enough for the bug.

Addtionally, ingest.New now instead of directly accepting a `kusto.Client`, now accepts an interface called `ingest.QueryClient` that contains all the relevant methods. This lets us mock the client, and might have more creative uses in the future.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -45,6 +45,8 @@ jobs:
           APP_KEY: ${{ secrets.APP_KEY }}
           AUTH_ID: ${{ secrets.AUTH_ID }}
           TEST_DATABASE: ${{ secrets.TEST_DATABASE }}
+          SECONDARY_ENGINE_CONNECTION_STRING: ${{ secrets.SECONDARY_ENGINE_CONNECTION_STRING }}
+          SECONDARY_DATABASE: ${{ secrets.SECONDARY_DATABASE }}
       - name: Publish Unit Test Results
         uses: EnricoMi/publish-unit-test-result-action@v1
         if: always()
--- a/kusto/ingest/ingest_test.go
+++ b/kusto/ingest/ingest_test.go
@@ -1,0 +1,143 @@
+package ingest
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Azure/azure-kusto-go/kusto"
+	"github.com/Azure/azure-kusto-go/kusto/data/table"
+	"github.com/Azure/azure-kusto-go/kusto/data/types"
+	"github.com/Azure/azure-kusto-go/kusto/ingest/internal/resources"
+)
+
+type mockClient struct {
+	endpoint string
+	auth     kusto.Authorization
+}
+
+func (m mockClient) Auth() kusto.Authorization {
+	return m.auth
+}
+
+func (m mockClient) Endpoint() string {
+	return m.endpoint
+}
+
+func (m mockClient) Query(context.Context, string, kusto.Stmt, ...kusto.QueryOption) (*kusto.RowIterator, error) {
+	panic("not implemented")
+}
+
+func (m mockClient) Mgmt(context.Context, string, kusto.Stmt, ...kusto.MgmtOption) (*kusto.RowIterator, error) {
+	rows, err := kusto.NewMockRows(table.Columns{
+		{
+			Name: "ResourceTypeName",
+			Type: types.String,
+		},
+		{
+			Name: "StorageRoot",
+			Type: types.String,
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+	iter := &kusto.RowIterator{}
+	err = iter.Mock(rows)
+	if err != nil {
+		return nil, err
+	}
+
+	return iter, nil
+}
+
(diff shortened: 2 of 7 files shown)


.github/workflows/go.yml
```diff
@@ -45,6 +45,8 @@ jobs:
           APP_KEY: ${{ secrets.APP_KEY }}
           AUTH_ID: ${{ secrets.AUTH_ID }}
           TEST_DATABASE: ${{ secrets.TEST_DATABASE }}
+          SECONDARY_ENGINE_CONNECTION_STRING: ${{ secrets.SECONDARY_ENGINE_CONNECTION_STRING }}
```
yogilad, line 48: Which engine is that?

kusto/ingest/ingest_test.go
```diff
@@ -0,0 +1,143 @@
...
+				},
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			mgrMap := make(map[*resources.Manager]bool)
+			for _, client := range test.clients {
+				mgr := client().mgr
```
yogilad, line 134: I don't think this test contributes much given the code you changed.
We do not check that new objects are actually new unless there is a caching logic which you clearly removed

kusto/ingest/internal/resources/resources.go
```diff
@@ -17,7 +17,7 @@ import (
 
 // mgmter is a private interface that allows us to write hermetic tests against the kusto.Client.Mgmt() method.
 type mgmter interface {
```
yogilad, line 19: mgmter?
Not very readable.
I get 'mgmt', what does 'er' stand for?

yogilad: Also, if your sole purpose if testing, why not use the below Query Client.

kusto/test/etoe/etoe_test.go
```diff
@@ -355,11 +355,21 @@ func TestFileIngestion(t *testing.T) {
 		panic(err)
 	}
 
+	secondaryClient, err := kusto.New(testConfig.SecondaryEndpoint, testConfig.Authorizer)
```
yogilad, line 358: This will make it hard to run tests locally.
Can we make these specific new tests optional?

kusto/test/etoe/etoe_test.go
```diff
@@ -490,6 +502,33 @@ func TestFileIngestion(t *testing.T) {
 			},
 			want: mockRows,
 		},
+		{
+			desc: "Ingestion to two different clusters",
+			src:  createCsvFileFromData(mockRows),
+			stmt: kusto.NewStmt("Logs | order by header_api_version asc"),
+			setup: func() error {
+				err := createIngestionTable(client, testConfig.Database)
```
yogilad, line 510: Do you delete these tables after?

.github/workflows/go.yml
AsafMah: In actions? https://sdke2etestsecondary.eastus.kusto.windows.net

kusto/ingest/internal/resources/resources.go
AsafMah: That's not new code of mine, Mgmter was before.

It's one of those annoying go convensions - if an interface has one method called X, the interface should be called Xer.

I could switch it to QueryClient I guess, but then I have to also change the tests that provide it a mgmter mock

kusto/test/etoe/etoe_test.go
AsafMah: Currently no. Also the tables don't have unique names per test.

These are both fixed in the StreamingIngest branch though, so once we'll merge it will be ok.

kusto/ingest/ingest_test.go
AsafMah: Yes I guess after removing the cache they don't do much,
but if we'll add a cache it will force us to fix them.

But your choice

.github/workflows/go.yml
AsafMah: Resolved offline

kusto/ingest/internal/resources/resources.go
AsafMah: Resolved offline

kusto/ingest/ingest_test.go
AsafMah: Resolved offline

kusto/test/etoe/etoe_test.go
AsafMah: Look if that's what you had in mind

yogilad (approved):

Issue #73, opened 2022-02-09
`TableFragmentType == DataReplace` is not correctly handled


w1ndy:
Queries may return table fragments of type `DataReplace` as per [docs](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/api/rest/response2#tablefragment), which are not correctly handled. The result set may contain duplicated entries.

w1ndy:
Close now since this issue is addressed by 9a620e9

PR #74, opened 2022-02-09
Expose `TableFragmentType` via a Replace flag

w1ndy:
#### Pull Request Description

This PR proposes a fix for #73 . Table fragments of type `DataReplace` should cause the existing result set to be replaced with new one; otherwise, the obtained query result may contain duplicated entries. To make users aware of this, this PR adds a flag called `Replace` in struct `Row` to indicate that the existing result set should be emptied before adding this row to the set.

---

#### Future Release Comment

**Breaking Changes:**
- The new `Replace` flag in struct `Row` should be handled to avoid duplicated entries in the result set. Adding the following code to handle the `Replace` flag in the `iter.Do` calls is recommended:

  ```golang
  var results []Result
  iter.Do(
	  func(row *table.Row) error {
		  // result := ...
		  if row.Replace {
			  results = results[:0]
		  }
		  results = append(results, result)
		  return nil
	  },
  )
  ```

**Fixes:**
- Fix duplicated entries caused by `TableFragmentType` (#73)

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ defer iter.Stop()
 err = iter.Do(
 	func(row *table.Row) error {
 		if row.Replace {
-			fmt.Println("---")	// Replace flag indicates that the query result should be cleared and replaced with this row
+			fmt.Println("---") // Replace flag indicates that the query result should be cleared and replaced with this row
 		}
 		fmt.Println(row) // As a convenience, printing a *table.Row will output csv
 		return nil
@@ -106,7 +106,7 @@ err = iter.Do(
 			return err
 		}
 		if row.Replace {
-			recs = recs[:0]
+			recs = recs[:0] // Replace flag indicates that the query result should be cleared and replaced with this row
 		}
 		recs = append(recs, rec)
 		return nil
--- a/README.md
+++ b/README.md
@@ -66,6 +66,9 @@ defer iter.Stop()
 // .Do() will call the function for every row in the table.
 err = iter.Do(
 	func(row *table.Row) error {
+		if row.Replace {
+			fmt.Println("---")	// Replace flag indicates that the query result should be cleared and replaced with this row
+		}
 		fmt.Println(row) // As a convenience, printing a *table.Row will output csv
 		return nil
 	},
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -102,6 +105,9 @@ err = iter.Do(
 		if err := row.ToStruct(&rec); err != nil {
 			return err
 		}
+		if row.Replace {
+			recs = recs[:0]
```
AsafMah, line 109: Should there also be a comment here?

AsafMah:
Overall looks good, just needs a test in statemachine_test.go

README.md
w1ndy: Yes, indeed, I've added a comment to this line

w1ndy:
@AsafMah Thanks for your review! I've added a simple test to statemachine_test.go. Could you help me review again?

kusto/doc.go
```diff
@@ -44,6 +44,9 @@ and accepts only string constants for arguments.
 	// .Do() will call the function for every row in the table.
 	err = iter.Do(
 		func(row *table.Row) error {
+			if row.Replace {
+				fmt.Println("---")	// Replace flag indicates that the query result should be cleared and replaced with this row
```
AsafMah, line 48:
```suggestion
				fmt.Println("---") // Replace flag indicates that the query result should be cleared and replaced with this row
```

AsafMah: to keep consistent with the other comments

kusto/statemachine_test.go
```diff
@@ -407,6 +407,7 @@ func TestProgressive(t *testing.T) {
 							value.Long{Value: 0, Valid: false},
 						},
 					},
+					TableFragmentType: "DataReplace",
```
AsafMah, line 410: Maybe add a test case to make sure that if we get multiple rows from a "DataReplace" table, the "Replace" flag should only be turned on for the first row

w1ndy:
@AsafMah Good suggestion, I have added the test

AsafMah (approved):
🚀

Issue #72, opened 2022-02-09
`row.Columns` populates column names instead of values


w1ndy:
Documentation says:

> The value of the kth column will be decoded into the kth argument to Columns

but actually the column names are decoded into the arguments.

https://github.com/Azure/azure-kusto-go/blob/b89f1684a83556dc7c03c161afacad5b29f6a167/kusto/data/table/table.go#L77-L104

w1ndy:
I looked into the test and found that `Columns` seems to retrieve column names by design... I wonder why this method is needed since there already exists similar ones like `ColumnNames` and `ColumnTypes`. Maybe we can change its behavior to retrieve values or replace it with a new method called `Values`?

element-of-surprise:
This is a leave over from my early design work that has somehow managed to stay in here.  Kudos for finding it.

When I was doing the first drafts I was looking at both database/sql,  spanner's and a few other APIs.  

I blame too many late nights for this, but this is really a completely wrong implementation.  The kind of thing I should have said, what the hell are you doing.......

What this should be doing is exactly what you said, grabbing the value from those positional locations in the row.

We should probably have this functionality, but this is now kind of a pickle. It can't be Values, we offer an attribute that provides the Values.  We can't change it, because that could cause someone some consequences.

I'm not the current maintainer, so I've sent this up the chain to see what they want to do.

w1ndy:
@element-of-surprise @yogilad Thanks for your effort! I think this issue may be addressed as follows:

1. Decouple the logic of `boolConvert`, `dateTimeConvert`, etc from `row.ToStruct` so they can be reused for any given `reflect.Type` and `reflect.Value`, instead of being tied to struct fields;
2. Add a new method (could be called `ExtractValues` or whatever) to read values from a row easily.

How do you like these changes? I can help implement them.

w1ndy:
Fixed by #75

PR #75, opened 2022-02-09
Refactor value converters and implement `ExtractValues` for `Row`

w1ndy:
#### Pull Request Description

This PR proposes a fix for #72 , which includes the following changes:

1. The logic of boolConvert, dateTimeConvert, etc are decoupled from `Row.ToStruct`. Now, each value type has its own Convert method, which transforms its value into a `reflect.Value`. All Convert methods are taken as is from the logic used by `Row.ToStruct`. Moreover, each value type is placed in its own go source file because the `value.go` file has too many lines.
2.  A new method called `ExtractValues` is added to struct Row. This method implements the intended functionality of `Columns`, which allows users to unpack the row's values into interfaces, enabling convenient and simplified access to these values.
3. Fix the docs of `Row.Columns`.

`ExtractValues` can be used like this:

```golang
var userId int64
var userName string
row.ExtractValues(&userId, &userName)
```

---

#### Future Release Comment

**Features:**
- `Row.ExtractValues` is added to easily unpack a row's values into a set of interfaces.


AsafMah:
It seems like that in the refactoring we lost this logic before each conversion (for each type it's slightly different of course):
``` 
val, ok := k.(value.Bool)
	if !ok {
		return fmt.Errorf("Column %s is type %s was trying to store a KustoValue type of %T", col.Name, col.Type, k)
	}
```

Is there a reason for removing this?

w1ndy:
> It seems like that in the refactoring we lost this logic before each conversion (for each type it's slightly different of course):
> 
> ```
> val, ok := k.(value.Bool)
> 	if !ok {
> 		return fmt.Errorf("Column %s is type %s was trying to store a KustoValue type of %T", col.Name, col.Type, k)
> 	}
> ```
> 
> Is there a reason for removing this?
> 
> Are there more changes to note?

To my best understanding, such a type conversion is not needed if we have a polymorphism method `Convert` on the interface `value.Kusto`. In the prior implementation, each convert method accepts different types of `value.Kusto`, so we'll have to manually do type casting before passing the values into these methods. Now, we can call the convert method on `value.Kusto` directly, and the implementation that will be executed depends on which type the value is. This also implicitly impose a constraint on each type inherited from `value.Kusto` where the type must have a Convert method to do such conversion.

This code snippet may help you understand the idea: https://go.dev/play/p/j4rZWDJrdCu

AsafMah:
Got it.

First of all, can you merge from main? there had been some changes and I want to resolve the conflicts early.

For now, can you expand the test you wrote to cover all types, just to make sure we didn't miss anything.

I also think we should check ExtractValue in etoe tests, I might add it myself to this PR.

AsafMah (approved):

Issue #81, opened 2022-03-16
Library panics at certain query failures


w1ndy:
Looks like ``v2.rawToOneAPIErr`` expects that errors are of type ``map[string]interface {}`` but actually an array of interfaces is returned...

```
panic: interface conversion: interface {} is []interface {}, not map[string]interface {}

goroutine 46 [running]:
github.com/Azure/azure-kusto-go/kusto/internal/frames/v2.rawToOneAPIErr({0xc000166000, 0x3bdd, 0x3e00}, 0xe800)
        /home/diweng/go/pkg/mod/github.com/!azure/azure-kusto-go@v0.5.2/kusto/internal/frames/v2/v2.go:217 +0x113
github.com/Azure/azure-kusto-go/kusto/internal/frames/v2.(*DataTable).UnmarshalRaw(0xc0001142d0, {0xc000166000, 0x3bdd, 0x3e00})
        /home/diweng/go/pkg/mod/github.com/!azure/azure-kusto-go@v0.5.2/kusto/internal/frames/v2/v2.go:61 +0x127
github.com/Azure/azure-kusto-go/kusto/internal/frames/v2.(*Decoder).decode(0xc000218380, {0x92a968, 0xc0002e2e40}, 0xc00005ae88)
        /home/diweng/go/pkg/mod/github.com/!azure/azure-kusto-go@v0.5.2/kusto/internal/frames/v2/decoder.go:117 +0x81f
github.com/Azure/azure-kusto-go/kusto/internal/frames/v2.(*Decoder).decodeFrames(0xc000218380, {0x92a968, 0xc0002e2e40}, 0x0)
        /home/diweng/go/pkg/mod/github.com/!azure/azure-kusto-go@v0.5.2/kusto/internal/frames/v2/decoder.go:83 +0x66
github.com/Azure/azure-kusto-go/kusto/internal/frames/v2.(*Decoder).Decode.func1()
        /home/diweng/go/pkg/mod/github.com/!azure/azure-kusto-go@v0.5.2/kusto/internal/frames/v2/decoder.go:60 +0x37e
created by github.com/Azure/azure-kusto-go/kusto/internal/frames/v2.(*Decoder).Decode
        /home/diweng/go/pkg/mod/github.com/!azure/azure-kusto-go@v0.5.2/kusto/internal/frames/v2/decoder.go:33 +0x1af
exit status 2
```

The query response is as follows
```json
{
    "FrameType": "DataTable",
    "TableId": 1,
[...]

Issue #42, opened 2021-02-16
Panic while type asserting FieldRows to map[string]interface{} in rawToOneAPIErr


1lann:
I got the following panic while trying to query data. Let me know on Teams or email if you need more specific details (my alias is jasch). It doesn't seem to always happen as I'm unable to reproduce it consistently. The `kusto.ResultsProgressiveDisable()` is being used on the query that causes the panic.

```
panic: interface conversion: interface {} is []interface {}, not map[string]interface {}

goroutine 210 [running]:
github.com/Azure/azure-kusto-go/kusto/internal/frames/v2.rawToOneAPIErr(0xc01ee00000, 0x43bb7d4, 0x7fffe00, 0x7d0000, 0xc0152cb4d0, 0x8b5320)
		/home/jason/.go/pkg/mod/github.com/!azure/azure-kusto-go@v0.3.1/kusto/internal/frames/v2/v2.go:217 +0x17f
github.com/Azure/azure-kusto-go/kusto/internal/frames/v2.(*DataTable).UnmarshalRaw(0xc0152cb4d0, 0xc01ee00000, 0x43bb7d4, 0x7fffe00, 0x0, 0x0)
		/home/jason/.go/pkg/mod/github.com/!azure/azure-kusto-go@v0.3.1/kusto/internal/frames/v2/v2.go:61 +0x14f
github.com/Azure/azure-kusto-go/kusto/internal/frames/v2.(*Decoder).decode(0xc0000fa040, 0x8bb868, 0xc009e57980, 0xc0129fe000, 0x0, 0x0)
		/home/jason/.go/pkg/mod/github.com/!azure/azure-kusto-go@v0.3.1/kusto/internal/frames/v2/decoder.go:117 +0xb1e
github.com/Azure/azure-kusto-go/kusto/internal/frames/v2.(*Decoder).decodeFrames(0xc0000fa040, 0x8bb868, 0xc009e57980, 0xc0129fe000)
		/home/jason/.go/pkg/mod/github.com/!azure/azure-kusto-go@v0.3.1/kusto/internal/frames/v2/decoder.go:83 +0x85
github.com/Azure/azure-kusto-go/kusto/internal/frames/v2.(*Decoder).Decode.func1(0x8b7ca0, 0xc0000fe840, 0xc0129fe000, 0xc0000fa040, 0x8bb868, 0xc009e57980)
[...]

PR #82, opened 2022-03-24
Fixed parsing panic when row has errors

AsafMah:
#### Pull Request Description

Fixes parsing when there is an error in the middle of the rows

---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- #81 #42


kusto/reader.go
```diff
@@ -195,38 +211,80 @@ func (r *RowIterator) Do(f func(r *table.Row) error) error {
...
+	if inlineErr != nil {
+		r.setError(inlineErr)
+		return nil, inlineErr
+	}
+	return row, err
+}
+
+// Next2 gets the next Row from the query. io.EOF is returned if there are no more entries in the output, or if a permanent error has occurred.
+// If an error within a table's rows occurs, inlineError will be set.
+// Errors may be returned, but iteration should continue until io.EOF is returned.
```
yogilad, line 257:
```suggestion
// NextRowOrError gets the next Row or service-side error from the query. 
// On partial success, inlineError will be set.
// Once finalError returns non-nil, all subsequent calls will return the same error.
// finalError will be set to io.EOF is when frame parsing completed with success or partial success (data + errors).
// if finalError is not io.EOF, reading the frame has resulted in a failure state (no data is expected). 
```

yogilad (approved):

PR #89, opened 2022-04-11
Update azblob

AsafMah:
#### Pull Request Description

Update azblob to its new library and API, from the old deprecated API.
Should help with memory leaks.

---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- Replaced the deprecated azblob library with the new one


kusto/ingest/internal/queued/queued.go
```diff
@@ -264,27 +269,29 @@ func CompleteFormatFromFileName(props *properties.All, from string) error {
...
+		return azblob.ContainerClient{}, errors.ES(
 			errors.OpFileIngest,
 			errors.KBlobstore,
 			"no Blob Storage container resources are defined, there is no container to upload to",
 		).SetNoRetry()
 	}
 
 	storageURI := mgrResources.Containers[rand.Intn(len(mgrResources.Containers))]
-	service, _ := url.Parse(fmt.Sprintf("https://%s.blob.core.windows.net?%s", storageURI.Account(), storageURI.SAS().Encode()))
+	serviceURL := fmt.Sprintf("https://%s.blob.core.windows.net?%s", storageURI.Account(), storageURI.SAS().Encode())
```
yogilad, line 287: Does this need any URL escaping?

kusto/ingest/internal/queued/queued.go
```diff
@@ -314,19 +321,19 @@ var nower = time.Now
 
 // localToBlob copies from a local to to an Azure Blobstore blob. It returns the URL of the Blob, the local file info and an
 // error if there was one.
-func (i *Ingestion) localToBlob(ctx context.Context, from string, to azblob.ContainerURL, props *properties.All) (azblob.BlockBlobURL, int64, error) {
+func (i *Ingestion) localToBlob(ctx context.Context, from string, to azblob.ContainerClient, props *properties.All) (string, int64, error) {
```
yogilad, line 324: nit: "to" is a bad name

yogilad (approved):
Some minor suggestions

Do you have a way to perf test / Test Telegarf for memory footprint?

kusto/ingest/internal/queued/queued.go
AsafMah: I don't think so, we don't accept that data from the user - we accept it straight from kusto, and we do escape the SAS (with .Encode)

kusto/ingest/internal/queued/queued.go
AsafMah: is `container` good?

yogilad (approved):

Issue #46, opened 2021-04-01
http or https proxy support


frcai:
Looks like azure-kusto-go doesn't respect environment HTTP_PROXY (or HTTPS_PROXY) when ingest data to Kusto. We have some customers need uploading data to Kusto but client doesn't have direct access to internet and all outgoing traffic needs go through outbound proxy.

yogilad:
Hi,

I'm not familiar with this environment and I don't see a reference to it in our Gold Standard SDK (C#).
This could be something the underlying http client libraries do on their own and Go does not, or it may not be supported there either.

You may be able to workaround the issue by setting a `netsh portproxy` rule.

element-of-surprise:
It doesn't.  This is because we use our own http.Client instead of the default client.

The best way to support this is to allow customers to pass there own client.   We'd simply provide an option that pass a type that satisfies this interface:

```go
type HTTPClient interface {
	// Do sends an HTTP request and returns an HTTP response.
	Do(req *http.Request) (*http.Response, error)

	// CloseIdleConnections closes any idle connections in a "keep-alive" state.
	CloseIdleConnections()
}
```
They could pass whatever they want (different HTTP client implementations, the default one, whatever)...

frcai:
Thanks Yochai and John. It'll be great to support the option John proposed. Do we have rough timeline to support this feature request?

frcai:
Looking code today and found default http client was used, the issue is that wrong value  (with line break) configured on HTTP_PORXY (I'm using secrectRef in K8S cluster).  After removing the line break, data can be uploaded successfully through proxy. 

// NewPipeline creates a new goroutine-safe Pipeline object from the slice of Factory objects and the specified options.
func NewPipeline(factories []Factory, o Options) Pipeline {
	if o.HTTPSender == nil {
		o.HTTPSender = newDefaultHTTPClientFactory()
	}
	if o.Log.Log == nil {
		o.Log.Log = func(LogLevel, string) {} // No-op logger
	}
	return &pipeline{factories: factories, options: o}
}

PR #91, opened 2022-04-15
Add support for custom http.Client

14leeyuchieh:
#### Pull Request Description

Users need to be able to specify/configure http settings like
proxy & custom retries, which are only avaiable to set by exposing
http.Client as something the user can customize. For example,
in enterprise boxes, deployments may not have access to internet
and need to go through squid proxy. Setting HTTPS_PROXY is invasive,
as it forces other http clients to disable proxy in transport or edit NO_PROXY
in such a way that specific dns suffix are disabled. Furthermore, even if HTTPS_PROXY 
is specified, some proxy requires users to specify certs/authorization header.

#### Future Release Comment
package `github.com/azure-kusto-go/kusto` supports `NewWithCustomHttp` allowing users to instantiate
a query Kusto client with http customizations.

**Breaking Changes:**
- None

**Features:**
- This addresses https://github.com/Azure/azure-kusto-go/issues/46 from a code perspective

**Fixes:**
- None

### DIFF
--- a/kusto/conn.go
+++ b/kusto/conn.go
@@ -42,8 +42,8 @@ type conn struct {
 	client                         *http.Client
 }
 
-// newConn returns a new conn object.
-func newConn(endpoint string, auth Authorization) (*conn, error) {
+// newConn returns a new conn object with an injected http.Client
+func newConn(endpoint string, auth Authorization, client *http.Client) (*conn, error) {
 	if !validURL.MatchString(endpoint) {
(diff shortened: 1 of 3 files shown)


AsafMah:
We usually prefer using Option for optional arguments instead of adding a new ctor, I'll make a new PR to reflect this.

AsafMah:
We prefer using options over creating a new ctor, so closed in favor of #92

PR #92, opened 2022-04-24
Added custom http option

AsafMah:
#### Pull Request Description

Users need to be able to specify/configure http settings like
proxy & custom retries, which are only avaiable to set by exposing
http.Client as something the user can customize. For example,
in enterprise boxes, deployments may not have access to internet
and need to go through squid proxy. Setting HTTPS_PROXY is invasive,
as it forces other http clients to disable proxy in transport or edit NO_PROXY
in such a way that specific dns suffix are disabled. Furthermore, even if HTTPS_PROXY 
is specified, some proxy requires users to specify certs/authorization header.

#### Future Release Comment
package `github.com/azure-kusto-go/kusto` supports `NewWithCustomHttp` allowing users to instantiate
a query Kusto client with http customizations.

**Breaking Changes:**
- None

**Features:**
- This addresses https://github.com/Azure/azure-kusto-go/issues/46 from a code perspective

**Fixes:**
- None

### DIFF
--- a/kusto/ingest/internal/conn/conn.go
+++ b/kusto/ingest/internal/conn/conn.go
@@ -44,7 +44,7 @@ type Conn struct {
 }
 
 // New returns a new Conn object.
-func New(endpoint string, auth kusto.Authorization) (*Conn, error) {
+func New(endpoint string, auth kusto.Authorization, client *http.Client) (*Conn, error) {
 	if !validURL.MatchString(endpoint) {
 		return nil, errors.ES(
 			errors.OpServConn,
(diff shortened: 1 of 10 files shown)


yogilad (commented):
How does this change cover ingestion flows i.e. how does it cover access to SAs?

AsafMah:
This is just for queries, users provide their own Auth, so they'll need to set whatever proxy or setting they have there.

AsafMah:
@yogilad Ok, I also made it so the http client will go to the ingestion flow

kusto/ingest/internal/conn/conn.go
```diff
@@ -44,7 +44,7 @@ type Conn struct {
 }
 
 // New returns a new Conn object.
-func New(endpoint string, auth kusto.Authorization) (*Conn, error) {
+func New(endpoint string, auth kusto.Authorization, client *http.Client) (*Conn, error) {
```
14leeyuchieh, line 47: Should this be called out as a breaking change? Technically, the constructor for the connection changed, right?

AsafMah: It's an internal file, so it's not breaking to change it.

yogilad (approved):

PR #98, opened 2022-04-27
Small refactoring to support 'HttpError'.

alonadam:


AsafMah:
MM you might want to try embedding Error inside HttpError, and then you can save (some of) the hassle

AsafMah (approved):
LGTM

PR #104, opened 2022-07-26
Feature/cloud info bugs

AsafMah:
#### Pull Request Description

---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None

### DIFF
--- a/kusto/cloudinfo.go
+++ b/kusto/cloudinfo.go
@@ -34,7 +34,6 @@ type cloudInfo struct {
 	firstPartyAuthorityUrl string
 }
 
-var doOnce sync.Once
 var defaultCloudInfo = &cloudInfo{
 	loginEndpoint:          getEnvOrDefault(defaultAuthEnvVarName, defaultPublicLoginUrl),
 	loginMfaRequired:       false,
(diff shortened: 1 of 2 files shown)


kusto/cloudinfo.go
```diff
@@ -0,0 +1,138 @@
...
+
+	if err != nil {
+		// TODO how do we log
+		errorToReturn = err
+	}
+	// metadata retrieval was successful
+	if metadataResponse.StatusCode == 200 {
+		// close once read
+		defer metadataResponse.Body.Close()
+		jsonBytes, resError := ioutil.ReadAll(metadataResponse.Body)
```
element-of-surprise, line 87: Let's make this way easiser:

```go
b, err := ioutil.ReadAll(resp.Body)
if err != nil {
  return CloudInfo{}, err // I'm assuming we do away with the pointer, so now we return the zero value
}

// This piece of code is suspect to me.  If you don't get a body, isn't that a server error.  And if it is
// and the result is always to just return defaultCloudInfo, then it seems we should do this differently.
// But I'm going to leave it in for the moment.
if len(b) == 0 {
  cloudInfoCache[kustoURL] = defaultCloudInfo
  return defaultCloudInfo
}

m := map[string]map[string]interface{}{} // This should really be a concrete type.
if err := json.Unmarshal(b, &m); err != nil { // You were ignoring the error here, which you shouldn't do.
  return CloudInfo{}, fmt.Errorf("metadata service returned bad JSON: %w", err)
}
...
```

I think you should have a concrete type I did not include above for this whole nested data, then you can just JSON marshal directly in.  This only requires one struct and making the CloudInfo fields public.

```go
type metaResp struct {
  AzureAD AzureADResp
}

type CloudInfo struct {
  LoginMFARequired bool `json: "LoginMfaRequired"`
  KustoClientAppID string `json: "KustoClientAppId"`
  KustoClientRedirectURI string `json: "KustoClientRedirectUri"`
  KustoServiceResourceID string `json: "KustoServiceResourceId"`
  FirstPartyAuthorityURL string `json: "FirstPartyAuthorityUrl"
}
```

Now you could just do this directly with:

```go 
md := metaResp{}
if err := json.Unmarshal(b, &md); err != nil {
  ...
}
```

kusto/cloudinfo.go
```diff
@@ -0,0 +1,138 @@
...
+		// the URL is not reachable , fallback to default
+		// For now as long not all proxies implement the metadata endpoint, if no endpoint exists return public cloud data
+		// TODO warn logging here
+		cloudInfoCache[kustoUrl] = defaultCloudInfo
+	} else {
+		// Some other HTTP error code here
+		errorToReturn = fmt.Errorf("retrieved error code %d when querying endpoint %s", metadataResponse.StatusCode, fullMetadataEndpoint)
+	}
+
+	if errorToReturn != nil {
```
element-of-surprise, line 126: This would no longer be needed.

kusto/cloudinfo.go
```diff
@@ -0,0 +1,138 @@
...
+			cloudInfoRetrieved := &cloudInfo{
+				loginEndpoint:          nestedMap["LoginEndpoint"].(string),
+				loginMfaRequired:       nestedMap["LoginMfaRequired"].(bool),
+				kustoClientAppId:       nestedMap["KustoClientAppId"].(string),
+				kustoClientRedirectUri: nestedMap["KustoClientRedirectUri"].(string),
+				kustoServiceResourceId: nestedMap["KustoServiceResourceId"].(string),
+				firstPartyAuthorityUrl: nestedMap["FirstPartyAuthorityUrl"].(string),
+			}
+			// Add this into the cache
+			cloudInfoCache[kustoUrl] = cloudInfoRetrieved
```
element-of-surprise, line 113: You'd should simply return after this:

```go
cloudInfoCache[kustoURL] = cloudInfoRetrieved
return cloudInfoRetrieved, nil
```

kusto/cloudinfo.go
```diff
@@ -0,0 +1,138 @@
...
+	kustoClientAppId:       defaultKustoClientAppId,
+	kustoClientRedirectUri: defaultRedirectUri,
+	kustoServiceResourceId: defaultKustoServiceResourceId,
+	firstPartyAuthorityUrl: defaultFirstPartyAuthorityUrl,
+}
+
+// cache to query it once per instance
+var cloudInfoCache map[string]*cloudInfo
+
+func RetrieveCloudInfoMetadata(kustoUrl string) (*cloudInfo, error) {
```
element-of-surprise, line 49: The signature should be:

```go
func RetrieveCloudInfoMetadata(ctx context.Context, kustoURL string) (CloudInfo, error)
```
If something contacts something else, it likely needs a context object.  If for no other reason that in the future you may want to attach telemetry with this.  If you don't do it now and you need it later, you have to write a second call or make a breaking change.

The other changes to the signature will be discussed below.

kusto/cloudinfo.go
```diff
@@ -0,0 +1,138 @@
+package kusto
```
element-of-surprise, line 1: This seems like a worthy addition, but there are some changes that need to be made to bring it up to the level it needs to be at.   There is a lot here to unpack here and I imagine it will take several go throughs to get to an end product.

kusto/cloudinfo_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+func TestRetrieveCloudInfoMetadataSuccessNoBody(t *testing.T) {
+	server := runServer(http.StatusOK, ``, true)
+	actualCloudInfo, err := RetrieveCloudInfoMetadata(server.URL)
+	server.Close()
+	assert.Nil(t, err)
+	assert.Equal(t, defaultCloudInfo, actualCloudInfo)
+}
+
+func runServer(statusCode int, payload string, isSuccess bool) *httptest.Server {
```
element-of-surprise, line 81: To combine the tests above into a single table driven test, you can expand this.

```go
type server struct {
	code    int
	payload []byte
	http    *httptest.Server
}

func newTestServ() *server {
	s := &server{}
	s.http = httptest.NewServer(s)
	return s
}

func (s *server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	log.Println("got here")
	defer log.Println("exited")
	w.WriteHeader(s.code)
	if s.code != 200 {
		w.Write(s.payload)
	}
}

func (s *server) urlStr() string {
	return s.http.URL
}

func (s *server) close() {
	s.http.Close()
}
```

Then you can use it in each iteration of the test by:

```go
         s := newTestServ()
	defer s.close()

        for _, test := range tests {
	  s.code = test.code
	  s.payload = test.payload
          ...

          resp, err := Get(s.urlStr())
          switch {
          case err != nil && !test.err:
            t.Errorf("TestGet(%s): got err == %s, want err == nil", test.desc, err)
            continue
          case err == nil && test.err:
            t.Errorf("TestGet(%s): got err == nil, want err != nil", test.desc)
            continue
          case err != nil:
            continue
          }
         // Do assert stuff here
        }
        }
```

kusto/cloudinfo_test.go
```diff
@@ -0,0 +1,93 @@
...
+				loginMfaRequired:       true,
+				kustoClientAppId:       "db662dc1-0cfe-4e1c-a843-19a68e65bxxx",
+				kustoClientRedirectUri: "https://microsoft/kustoclient",
+				kustoServiceResourceId: "https://kusto.dev.kusto.windows.net",
+				firstPartyAuthorityUrl: "https://login.microsoftonline.com/f8cdef31-a31e-4b4a-93e4-5f571e912xxx",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
element-of-surprise, line 44: You can do this sub test runner, but not a lot of reason to do it here.  See comment on runServer() about how to combine all these tests into a single table driven test.

kusto/cloudinfo.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return CloudInfo{}, err
+	}
+
+	// Handle internal server error as a special case and return as an error (to be consistent with other SDK's)
+	if resp.StatusCode >= http.StatusInternalServerError {
+		return CloudInfo{}, fmt.Errorf("error %s when querying endpoint %s", resp.Status, u.String())
+	}
+
+	//  return fmt.Errorf("error %s when querying endpoint %s", metadataResponse.StatusCode, fullMetadataEndpoint)
```
AsafMah, line 89: Remove

kusto/cloudinfo.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Covers scenarios of 200/OK with no body or a 404 where there is no body
+	if len(b) == 0 {
+		cloudInfoCache[kustoURL] = defaultCloudInfo
+		return defaultCloudInfo, nil
+	}
+
+	md := metaResp{}
+
+	if err := json.Unmarshal(b, &md); err != nil {
```
AsafMah, line 106: What happens if not all of the properties exist? I think we should error, so make sure that what happens with a test

AsafMah (commented):
Looks good - small comments

element-of-surprise:
I'm OOO, ping me Monday and I'll have a look.

Cheers,

John Doak
Manager, Layer 1 Reliability Engineering
***@***.***
________________________________

PR #105, opened 2022-08-16
Feature/msa lmigration

asaharn:
#### Pull Request Description

---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None


AsafMah:
Is this ready to review? if so please put it out of draft.

Also - the build fails because of formatting, please run `go fmt` on your code, and make sure all the tests pass.

asaharn:
@AsafMah Raised the draft PR since there are few test cases need to be added to it. By the time the main files can be reviewed.

kusto/TokenProvider.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+type TokenProvider struct {
+	accessToken azcore.AccessToken //holds the received AccessToken for respective credential
+}
+
+func (tkp TokenProvider) getToken() azcore.AccessToken {
+	return tkp.accessToken
```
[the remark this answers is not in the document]
yogilad: Also, where do we leverage cloud info?
How do we pass the resource Id to the token?

You should follow a python example here if this wrapper is really needed

kusto/conn.go
```diff
@@ -159,6 +165,10 @@ func (c *conn) execute(ctx context.Context, execType int, db string, query Stmt,
 		return execResp{}, errors.ES(op, errors.KInternal, "internal error: did not understand the type of execType: %d", execType)
 	}
 
+	if c.tokenProvider != (TokenProvider{}) {
```
yogilad, line 168: It's a bit odd both auth blocks are not conditional one on the other

kusto/kcsb.go
```diff
@@ -0,0 +1,217 @@
...
+	clientIdEnvVariable     = "AZURE_CLIENT_ID"
+	clientSecretEnvVariable = "AZURE_CLIENT_SECRET"
+)
+
+// params mapping
+const (
+	tenantIDStr      string = "TenantID"
+	clientIDStr      string = "ClientID"
+	clientSecretStr  string = "ClientSecret"
+	appCertStr       string = "ApplicationCertificates"
```
yogilad, line 34: plural?
```suggestion
	appCertStr       string = "ApplicationCertificate"
```

kusto/kcsb.go
```diff
@@ -0,0 +1,217 @@
...
+	appCertKeyStr    string = "ApplicationCertificateKey"
+	usernameStr      string = "Username"
+	userPasswordStr  string = "UserPassword"
+	appTokenStr      string = "ApplicationToken"
+	clientOptionsStr string = "ClientOptions"
+	sendCertChainStr string = "SendCertificateChain"
+	managedIDStr     string = "ManagedIdentityID"
```
yogilad, line 41: We support this capability (mi from connection string) due to security reasons.
Same for certificate subject name and certificate thumbprint

kusto/kcsb.go
```diff
@@ -0,0 +1,217 @@
...
+	azCliAuth      string = "AzCLI"
+	managedIDAuth  string = "ManagedIdentity"
+	clientCredAuth string = "ClientCredentials"
+	appCertAuth    string = "ApplicationCetrifiate"
+	unamePassAuth  string = "UsernamePassword"
+	appTokenAuth   string = "ApplicationToken"
+)
+
+// Method to be used for generating TokenCredential
+func (kcsb *ConnectionStringBuilder) getTokenProvider(ctx context.Context) (*TokenProvider, error) {
```
yogilad, line 56: This is an entirely new concept for KCSB (at least compared to C# and Python)
Let's go with the model already established in the past

kusto/kusto.go
```diff
@@ -36,13 +36,17 @@ type Authorization struct {
...
+	// Token provider that can be used to get the access token.
+	tokenProvider TokenProvider
 }
 
 // Validate validates the Authorization object against the endpoint an preps it for use.
 // For internal use only.
 func (a *Authorization) Validate(endpoint string) error {
 	const rescField = "Resource"
-
+	if a.tokenProvider != (TokenProvider{}) {
```
yogilad, line 47: Q: What does this check do actually ?

kusto/kusto.go
```diff
@@ -119,7 +123,15 @@ type Client struct {
 type Option func(c *Client)
 
 // New returns a new Client. endpoint is the Kusto endpoint to use, example: https://somename.westus.kusto.windows.net .
-func New(endpoint string, auth Authorization, options ...Option) (*Client, error) {
```
yogilad, line 122: Doesn't this change effectively remove the ability to use ADAL Authorization?
If so, why keep it in the code?

kusto/kcsb.go
```diff
@@ -0,0 +1,217 @@
...
+
+			if !(tnok && clok && csok) {
+				return nil, errors.New("Error : Couldn't get token provider due to insufficient parameters")
+			}
+
+			ccred, err := azidentity.NewClientSecretCredential(tenantID.(string), clientId.(string), clientSec.(string), nil)
+			if err != nil {
+				return nil, fmt.Errorf("Error : Couldn't get client credentiels. Error: %s", err)
+			}
+			tkp.accessToken, err = ccred.GetToken(ctx, policy.TokenRequestOptions{Scopes: scopes})
```
yogilad, line 79: You actually want to return the cred, not the token

yogilad (changes requested):

kusto/kusto.go
asaharn: @yogilad Since we cannot compare the structs with nil in Go. We have to compare it with an empty struct of same type to check if its unset.

kusto/kusto.go
asaharn: Proceeding to remove the corresponding code...

kusto/kusto.go
AsafMah: Maybe hold a pointer that can be nil?

asaharn: @AsafMah Yeah agree to it. But we needed a solid reason to go for a pointer-based approach.   Instead, we will be using an isInitialized flag in the struct to check if the struct is created or not...

asaharn: one reason could be, if we make tokenProvider of pointer type, the reference would be used, hence making it more reliable in cases where any modifications occur in tokenProvider.

kusto/kcsb.go
```diff
@@ -11,191 +10,351 @@ import (
 )
 
 type connectionStringBuilder struct {
-	clusterURI string
-	authType   string
-	authParams map[string]interface{}
-	cloudInfo  CloudInfo
+	DataSource                       string `json:"DataSource"`
```
ag-ramachandran, line 13: The struct is private and the members are public. Let's make it consistent (private vars may be)

kusto/kcsb.go
```diff
@@ -11,191 +10,351 @@ import (
...
-const (
-	envAuth        string = "EnvironmentVars"
-	azCliAuth      string = "AzCLI"
-	managedIDAuth  string = "ManagedIdentity"
-	clientCredAuth string = "ClientCredentials"
-	appCertAuth    string = "ApplicationCetrifiate"
-	unamePassAuth  string = "UsernamePassword"
-	appTokenAuth   string = "ApplicationToken"
-)
+func assertIfEmpty(key string, value string) {
```
ag-ramachandran, line 52: rename it a little better requireNonEmpty() may be?

kusto/kcsb.go
```diff
@@ -11,191 +10,351 @@ import (
...
+		if tofind == s {
+			return true
+		}
+	}
+	return false
+}
+
+func assignValue(kcsb *connectionStringBuilder, rawKey string, value string) {
+	rawKey = strings.ToLower(strings.Trim(rawKey, " "))
+	if contains([]string{"datasource", "data source", "addr", "address", "network address", "server"}, rawKey) {
```
ag-ramachandran, line 69: Can this be simplified ? Essentially we have a KV string that is parsed to a JSON ?
https://stackoverflow.com/questions/35663892/parse-string-into-map-golang

kusto/kcsb.go
```diff
@@ -11,191 +10,351 @@ import (
...
+	}
+	if contains([]string{"interactive login", "interactivelogin"}, rawKey) {
+		kcsb.InteractiveLogin = getBoolValue(value)
+	}
+	if contains([]string{"domain hint", "domainhint"}, rawKey) {
+		kcsb.RedirectURL = value
+	}
+}
+
+func getBoolValue(val string) bool {
```
ag-ramachandran, line 119: equivalent of         boolValue, err := strconv.ParseBool(v) ? Do we need a method for this ?

kusto/kcsb.go
```diff
@@ -11,191 +10,351 @@ import (
...
+		assignValue(&kcsb, kvparr[0], val)
+
+	}
+
+	return kcsb
+}
+
+//Creates a KustoConnection string builder that will authenticate with AAD user name and password.
+func (kcsb connectionStringBuilder) WithAadUserPassAuth(uname string, pswrd string, authorityID string) connectionStringBuilder {
+	assertIfEmpty(dataSource, kcsb.DataSource)
```
ag-ramachandran, line 155: In hindsight , will a variadic function that takes all the parameters for the null check at once be simpler than checking each parameter ?

ag-ramachandran: Would it make sense to have *connectionStringBuilder and changing the attributes / adding attributes as opposed to taking a value and returning the value again with the mutation ?

kusto/TokenProvider.go
```diff
@@ -0,0 +1,29 @@
+package kusto
```
AsafMah, line 1: File Name

kusto/kcsb.go
AsafMah: There are multiple keys that can fit each item. I think using a map is correct here.

kusto/kcsb.go
```diff
@@ -0,0 +1,362 @@
...
+		kcsb.ClientOptions = options
+	}
+	return kcsb
+}
+
+// Method to be used for generating TokenCredential
+func (kcsb *connectionStringBuilder) getTokenProvider(ctx context.Context) (*tokenProvider, error) {
+	tkp := &tokenProvider{}
+	//Fetches cloud meta data
+	fetchedCI, cierr := GetMetadata(context.Background(), kcsb.DataSource)
```
AsafMah, line 257: Having the metadata fetch here is something we try to avoid in the other sdks.
Because it means that just creating a client is a blocking operation that waits on a web request, plus we don't need to do it for all providers - like the constant token one

kusto/kcsb.go
```diff
@@ -0,0 +1,362 @@
...
+		opts := &azidentity.AzureCLICredentialOptions{}
+		opts.TenantID = kcsb.AuthorityId
+		cred, err := azidentity.NewAzureCLICredential(opts)
+		if err != nil {
+			return nil, fmt.Errorf("Error : Couldn't retrieve client credentiels using Azure CLI: %s", err)
+		}
+		tkp.tokenCred = cred
+	} else {
+		//env variables based auth
+		opts := &azidentity.EnvironmentCredentialOptions{}
```
AsafMah, line 347: No way for the user to provide a kusto credential?

kusto/kcsb.go
```diff
@@ -0,0 +1,362 @@
...
+	password                         string = "Password"
+	applicationClientId              string = "ApplicationClientId"
+	applicationKey                   string = "ApplicationKey"
+	applicationCertificate           string = "ApplicationCertificate"
+	applicationCertificateThumbprint string = "ApplicationCertificateThumbprint"
+	sendCertificateChain             string = "SendCertificateChain"
+	authorityId                      string = "AuthorityId"
+	applicationToken                 string = "ApplicationToken"
+	userToken                        string = "UserToken"
+	msiAuth                          string = "MSIAuthentication"
```
AsafMah, line 45: Half of these are unsused

kusto/kcsb.go
```diff
@@ -0,0 +1,362 @@
...
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+)
+
+type connectionStringBuilder struct {
```
AsafMah, line 12: Is there a reason it needs to be serialized into json?

kusto/kcsb.go
```diff
@@ -0,0 +1,362 @@
...
+func (kcsb connectionStringBuilder) AttachClientOptions(options *azcore.ClientOptions) connectionStringBuilder {
+	assertIfEmpty(dataSource, kcsb.DataSource)
+	if options == nil {
+		kcsb.ClientOptions = options
+	}
+	return kcsb
+}
+
+// Method to be used for generating TokenCredential
+func (kcsb *connectionStringBuilder) getTokenProvider(ctx context.Context) (*tokenProvider, error) {
```
AsafMah, line 254: ctx is unused

kusto/kcsb.go
```diff
@@ -0,0 +1,362 @@
...
+	tkp := &tokenProvider{}
+	//Fetches cloud meta data
+	fetchedCI, cierr := GetMetadata(context.Background(), kcsb.DataSource)
+	if cierr != nil {
+		return nil, cierr
+	}
+	//Update resource URI if MFA enabled
+	resourceURI := fetchedCI.KustoServiceResourceID
```
AsafMah, line 262: Move to a method, maybe inside metadata?

kusto/kusto_examples_test.go
```diff
@@ -26,11 +26,8 @@ func Example_simple() {
 		CollectionTime time.Time
 	}
 
-	authorizer := Authorization{
-		Config: auth.NewClientCredentialsConfig("clientID", "clientSecret", "tenantID"),
-	}
-
-	client, err := New("endpoint", authorizer)
+	kcsb := GetConnectionStringBuilder("endpoint;tenantid=tenentID;appclientid=clientID;appkey=clientSecret")
```
AsafMah, line 29: I think the correct way to build a csb is with the builder methods, not to parse one

kusto/ingest/ingest_example_test.go
```diff
@@ -6,26 +6,14 @@ import (
...
 
-	authConfig := auth.NewClientCredentialsConfig("clientID", "clientSecret", "tenantID")
-	/*
-		Alteratively, you could so something like:
-		authorizer, err := auth.NewMSIConfig().Authorizer()
-		or
-		authorizer, err := auth.NewAuthorizerFromEnvironment()
-		or
-		auth.New...()
+	kcsb := kusto.GetConnectionStringBuilder(`endpoint;tenantid=tenantID;appclientid=clientID;appkey=clientSecret`)
```
AsafMah, line 14: Same here - show the simplest method to create

kusto/test/etoe/etoe_test.go
```diff
@@ -105,7 +105,9 @@ func TestQueries(t *testing.T) {
 	_, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	client, err := kusto.New(testConfig.Endpoint, testConfig.Authorizer)
+	kcsb := kusto.GetConnectionStringBuilder(fmt.Sprintf(`%s;tenantid=%s;appclientid=%s;appkey=%s`, testConfig.Endpoint, testConfig.TenantID, testConfig.ClientID, testConfig.ClientSecret))
```
AsafMah, line 108: That's really weird, why not use the ctor here?

kusto/conn.go
```diff
@@ -204,9 +204,9 @@ func (c *conn) execute(ctx context.Context, execType int, db string, query Stmt,
 }
 
 func (c *conn) Close() error {
-	if closer, ok := c.auth.(io.Closer); ok {
+	/*if closer, ok := c.auth.(io.Closer); ok {
```
AsafMah, line 207: ?

kusto/kcsb.go
AsafMah: Because it doesn't take a pointer it doesn't mutate the provided value. It would have done it if it was a pointer.

Maybe these need to be ctors anyway, since they are mutually exclusive.

kusto/kcsb.go
AsafMah, line 254: mixing pointer and non pointer functions is confusing.

Run a linter on the code to spot things like this.

AsafMah (changes requested):
Some comments

kusto/conn.go
asaharn: Since the auth for conn struct is removed so we had to remove this too.

kusto/kcsb.go
asaharn: As per discussion with Yochai, it seems the Kusto credential is just used for C# code, that too internal process. It doesn't make sense to be present in SDKs

kusto/kcsb.go
```diff
@@ -0,0 +1,354 @@
...
+	case managedServiceIdentity:
+		kcsb.managedServiceIdentity = value
+	case azCli:
+		bval, _ := strconv.ParseBool(value)
+		kcsb.azcli = bval
+	case interactiveLogin:
+		bval, _ := strconv.ParseBool(value)
+		kcsb.interactiveLogin = bval
+	case domainHint:
+		kcsb.redirectURL = value
```
yogilad, line 115: Add a default and report and error

kusto/kcsb.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+/*
+   Creates new KustoConnectionStringBuilder.
+   Params takes kusto connection string connStr: string.  Kusto connection string should be of the format:
+   https://<clusterName>.kusto.windows.net;AAD User ID="user@microsoft.com";Password=P@ssWord
+   For more information please look at:
+   https://docs.microsoft.com/azure/data-explorer/kusto/api/connection-strings/kusto
+*/
+func GetConnectionStringBuilder(connStr string) connectionStringBuilder {
```
yogilad, line 126: Why not use New()

kusto/kcsb.go
```diff
@@ -0,0 +1,354 @@
...
+	kcsb := connectionStringBuilder{}
+	if isEmpty(connStr) {
+		panic("Error : Connection string cannot be empty")
+	}
+
+	if !strings.Contains(strings.Split(connStr, ";")[0], "=") {
```
yogilad, line 132: call split once

kusto/kcsb.go
```diff
@@ -0,0 +1,354 @@
...
+
+	return kcsb
+}
+
+//Creates a KustoConnection string builder that will authenticate with AAD user name and password.
+func (kcsb connectionStringBuilder) WithAadUserPassAuth(uname string, pswrd string, authorityID string) connectionStringBuilder {
+	requireNonEmpty(dataSource, kcsb.dataSource)
+	requireNonEmpty(aadUserId, uname)
+	requireNonEmpty(password, pswrd)
+	kcsb.aadUserID = uname
```
yogilad, line 154: C# SDK rightfully clears all auth properties before assigning anything in WithXXX.
This means calling With resets previous auth options whether they came from the string or from a previous call to With

kusto/kcsb.go
```diff
@@ -0,0 +1,354 @@
...
+	kcsb.password = pswrd
+	kcsb.authorityId = authorityID
+	return kcsb
+}
+
+//Creates a KustoConnection string builder that will authenticate with AAD user token
+func (kcsb connectionStringBuilder) WitAadUserToken(userTkn string) connectionStringBuilder {
```
yogilad, line 161: userToken (avoid the abbreviation)

kusto/kcsb.go
```diff
@@ -0,0 +1,354 @@
...
+	kcsb.azcli = true
+	return kcsb
+}
+
+/*
+Creates a KustoConnection string builder that will authenticate with AAD application, using
+  an application token obtained from a Microsoft Service Identity endpoint. An optional user
+  assigned application ID can be added to the token.
+*/
+func (kcsb connectionStringBuilder) WithManagedServiceID(clientID string, resId string) connectionStringBuilder {
```
yogilad, line 214: Some pointers here since python is a little out of date.
1) only client id is needed - res id is not supported
2) if client id is not provided to Az identity it does system assigned auth
3) we prefer to have to API to avoid confusion - With System Managed Identity & With User Assigned Managed Identity

kusto/kcsb.go
```diff
@@ -0,0 +1,354 @@
...
+		kcsb.authorityId = authorityID
+	}
+	if !isEmpty(redirectURL) {
+		kcsb.redirectURL = redirectURL
+	}
+	kcsb.interactiveLogin = true
+	return kcsb
+}
+
+func (kcsb connectionStringBuilder) AttachClientOptions(options *azcore.ClientOptions) connectionStringBuilder {
```
yogilad, line 240: What's the purpose of client options?

kusto/kcsb.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+// Method to be used for generating TokenCredential
+func (kcsb connectionStringBuilder) getTokenProvider() (*tokenProvider, error) {
+	tkp := &tokenProvider{}
+	tkp.dataSource = kcsb.dataSource
+	if kcsb.interactiveLogin {
+		inOps := &azidentity.InteractiveBrowserCredentialOptions{}
+
+		if !isEmpty(kcsb.applicationClientId) {
```
yogilad, line 255: You need to fill this out with defaults from cloud settings.

kusto/kcsb.go
```diff
@@ -0,0 +1,354 @@
...
+			inOps.TenantID = kcsb.authorityId
+		}
+		if kcsb.clientOptions != nil {
+			inOps.ClientOptions = *kcsb.clientOptions
+		}
+		cred, err := azidentity.NewInteractiveBrowserCredential(inOps)
+		if err != nil {
+			return nil, fmt.Errorf("Error : Couldn't retrieve client credentiels using Interactive Login. Error: %s", err)
+		}
+		tkp.tokenCred = cred
```
yogilad, line 268: Return and then you can skip the else

kusto/kcsb.go
yogilad: Probably in a lot of places below

kusto/kcsb.go
```diff
@@ -0,0 +1,354 @@
...
+		inOps := &azidentity.InteractiveBrowserCredentialOptions{}
+
+		if !isEmpty(kcsb.applicationClientId) {
+			inOps.ClientID = kcsb.applicationClientId
+		}
+		if !isEmpty(kcsb.authorityId) {
+			inOps.TenantID = kcsb.authorityId
+		}
+		if kcsb.clientOptions != nil {
+			inOps.ClientOptions = *kcsb.clientOptions
```
yogilad, line 262: It's an odd line of code.
You do all the work of filling inOps only to override it.
I'm not sure why that member ops are needed for anyway.

kusto/tokenprovider.go
```diff
@@ -0,0 +1,52 @@
...
+			//Update resource URI if MFA enabled
+			resourceURI := fetchedCI.KustoServiceResourceID
+			if fetchedCI.LoginMfaRequired {
+				resourceURI = strings.Replace(resourceURI, ".kusto.", ".kustomfa.", 1)
+			}
+			tkp.scopes = []string{fmt.Sprintf("%s/.default", resourceURI)}
+		}
+		return tkp.tokenCred.GetToken(ctx, policy.TokenRequestOptions{Scopes: tkp.scopes})
+		
+	} else {
```
yogilad, line 44: We advise reducing scopes.
If you have a condition you can terminate the function in, return from the condition instead of if elseing, e.g.

```
x  int
if (a) {
 x = 1
} else if (b) {
x = 2
}  else {
c =3
}

return x

// should be 
if (a) {
return 1)
}

if (b) {
return 2
}

return 3
```

yogilad: Here for instance it should be (pseudo code)

```
if (has token string) {
return string
}

if (! token credential} {
return error
}

if (! cloud init) {
init cloud
}

retun get token 
```

kusto/kusto.go
```diff
@@ -28,80 +24,16 @@ type queryer interface {
...
-	// this instead of Authorizer, as we will automatically set the Resource ID with the endpoint passed.
-	Config auth.AuthorizerConfig
+	// Token provider that can be used to get the access token.
+	tokenProvider tokenProvider
 }
 
 // Validate validates the Authorization object against the endpoint an preps it for use.
 // For internal use only.
-func (a *Authorization) Validate(endpoint string) error {
+func (a *Authorization) Validate() error {
```
yogilad, line 33: We support no-auth mode for some cases (dev mostly).
Not sure if other SDKs enforce this.

@AsafMah  WDYT?

kusto/kcsb.go
asaharn: `Kusto.New()` is used for obtaining the `Client`. 
`Kusto.GetConnectionStringBuilder()` becomes better way. 
Either we go for whole new package name `Connectionstring`. So the `ConnectionString.New()` can be used..?

kusto/kcsb.go
asaharn: image
Sure @yogilad ! but would like to mention the point that `azidentity` have this provision to provide `resourceId` for `managedId`. If the plan is to not support `resourseId`, then it can be removed.

kusto/kcsb.go
asaharn: Basically [`ClientOptions`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore#ClientOptions) contains optional settings for a client's pipeline. Like cloud.Configuration, logging policy, retry policy, telemetry policy. 
Have added this to provide flexibility/ability to so that `ClientOptions` can be attached.

kusto/kcsb.go
asaharn: Have kept this way because, if the mentioned values are not present. `InteractiveBrowserCredentialOptions` will Defaults to the ID of the Azure development application.

kusto/cloudinfo.go
```diff
@@ -0,0 +1,119 @@
...
+// information needed for connection string builder to provide all the requisite information
+
+const (
+	metadataPath                  = "v1/rest/auth/metadata"
+	defaultAuthEnvVarName         = "AadAuthorityUri"
+	defaultKustoClientAppId       = "db662dc1-0cfe-4e1c-a843-19a68e65be58"
+	defaultPublicLoginUrl         = "https://login.microsoftonline.com"
+	defaultRedirectUri            = "https://microsoft/kustoclient"
+	defaultKustoServiceResourceId = "https://kusto.kusto.windows.net"
+	defaultFirstPartyAuthorityUrl = "https://login.microsoftonline.com/f8cdef31-a31e-4b4a-93e4-5f571e91255a"
```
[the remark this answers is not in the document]
AsafMah: It's URL in all other SDKs, I'm closing

kusto/cloudinfo.go
```diff
@@ -0,0 +1,119 @@
...
+
+	if err := json.Unmarshal(b, &md); err != nil {
+		return CloudInfo{}, err
+	}
+	// this should be set in the map by now
+	cloudInfoCache[kustoURL] = md.AzureAD
+	return md.AzureAD, nil
+}
+
+func getEnvOrDefault(key, fallback string) string {
```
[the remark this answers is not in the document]
AsafMah: Yes, we have this same thing in python.
To keep parity let's keep it

kusto/kcsb.go
AsafMah: Moreso, because kcsb is a big struct, you'd want to use a pointer in all of its methods, unless there is a good reason not too.

Also check other classes and make sure you don't mix pointers and non-pointers

kusto/kcsb.go
AsafMah: Remove resourceId entirely, we don't support it.
And like yochai said, make a method
`WithUserManagedIdentity` that accepts a clientID
And 
`WithSystemManagedIdentity` that doesn't

kusto/kusto.go
AsafMah: All other sdks have no auth mode, we should have it here too.

kusto/kcsb.go
AsafMah: We don't put the defaults from cloudinfo in any of the other SDKs. We put in the correct values when we fetch it, and that is what happens here, so I'm closing.

kusto/kcsb.go
AsafMah: TThis function sometimes early returns and sometimes doesn't. Please make it consistantly early return and then you can get rid of elses and indentations

kusto/kcsb.go
AsafMah: I don't see anything weird here, so I assume it was solved.

kusto/conn.go
```diff
@@ -181,6 +182,12 @@ func (c *conn) doRequest(ctx context.Context, execType int, db string, query Stm
 		return 0, nil, nil, nil, errors.ES(op, errors.KInternal, "internal error: did not understand the type of execType: %d", execType)
 	}
 
+	token, tokenType, tkerr := c.tokenProvider.AcquireToken(ctx)
```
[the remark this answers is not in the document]
AsafMah: We do have it in all of the SDKs, and part of why the tests failed is that we didn't have it here. I thought he added it?

kusto/kcsb.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+/*
+Creates new KustoConnectionStringBuilder.
+Params takes kusto connection string connStr: string.  Kusto connection string should be of the format:
+https://<clusterName>.kusto.windows.net;AAD User ID="user@microsoft.com";Password=P@ssWord
+For more information please look at:
+https://docs.microsoft.com/azure/data-explorer/kusto/api/connection-strings/kusto
+*/
+func GetConnectionStringBuilder(connStr string) *ConnectionStringBuilder {
```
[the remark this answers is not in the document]
asaharn: @yogilad Kcsb is in Kusto package, and the `New(...)` function is already being used. Should we create a new package just for `ConnectionStringBuilder` ? So that we can achieve ConnectionStringBuilder.New(...)

Intentions here is to use like:
* _If created with just string keywords_
 kusto.GetConnectionStringBuilder(...) 


* _If created using dedicated With*** apis_
{kcsb obj}.With***(...)

kusto/kcsb.go
```diff
@@ -0,0 +1,428 @@
...
+	kcsb.MsiAuthentication = false
+	kcsb.ManagedServiceIdentity = ""
+	kcsb.InteractiveLogin = false
+	kcsb.RedirectURL = ""
+	kcsb.ClientOptions = nil
+}
+
+// Creates a KustoConnection string builder that will authenticate with AAD user name and password.
+func (kcsb *ConnectionStringBuilder) WithAadUserPassAuth(uname string, pswrd string, authorityID string) *ConnectionStringBuilder {
+	requireNonEmpty(dataSource, kcsb.DataSource)
```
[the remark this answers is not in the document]
asaharn: This will pre check that some data source was defined in the passed Kcsb obj.

kusto/kcsb.go
```diff
@@ -0,0 +1,428 @@
...
+				inOps.TenantID = kcsb.AuthorityId
+			}
+			if kcsb.ClientOptions != nil {
+				inOps.ClientOptions = *kcsb.ClientOptions
+			}
+			cred, err := azidentity.NewInteractiveBrowserCredential(inOps)
+			if err != nil {
+				return nil, fmt.Errorf("Error : Couldn't retrieve client credentiels using Interactive Login. Error: %s", err)
+			}
+			tkp.tokenCred = cred
```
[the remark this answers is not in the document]
asaharn: Can be done that way but it would be inconsistent with the cases like custom userToken/appToken, where the tokenCred won't be present

kusto/kcsb.go
yogilad: @AsafMah WDYT?
New package or change the API name to **New**ConnectionStringBuilder?

kusto/kcsb.go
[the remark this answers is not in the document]
yogilad: I see I made this comment earlier (sorry).
If it stays in the same module, it should be **New**ConnectionStringBuilder.

kusto/conn.go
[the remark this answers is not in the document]
asaharn: @AsafMah @yogilad Yes, it is added here as well. The tests failed for internal/Conn.go

kusto/kusto.go
```diff
@@ -29,80 +25,16 @@ type queryer interface {
 // Authorization provides the ADAL authorizer needed to access the resource. You can set Authorizer or
 // Config, but not both.
 type Authorization struct {
```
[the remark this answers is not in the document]
asaharn: This could be done. But we kept it out of the scope this time to make the PR more focused and less complex.

kusto/test/etoe/etoe_env.go
```diff
@@ -86,17 +80,5 @@ func init() {
 		return
 	}
 
-	if testConfig.ClientID == "" {
```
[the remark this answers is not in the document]
asaharn: Added the functionality back.

kusto/cloudinfo.go
```diff
@@ -49,18 +49,14 @@ var defaultCloudInfo = CloudInfo{
...
 
 // cache to query it once per instance
-var cloudInfoCache = map[string]CloudInfo{}
-var lock = &sync.Mutex{}
+var cloudInfoCache sync.Map
 
 func GetMetadata(ctx context.Context, kustoUri string) (CloudInfo, error) {
 	// retrieve &return if exists
-	cachedCloudInfo, ok := cloudInfoCache[kustoUri]
+	anyCachedCloudInfo, ok := cloudInfoCache.Load(kustoUri)
```
yogilad, line 56:
```suggestion
	clusterCachedCloudInfo, ok := cloudInfoCache.Load(kustoUri)
```

kusto/kcsb.go
```diff
@@ -265,50 +257,68 @@ func (kcsb *ConnectionStringBuilder) WithSystemManagedIdentity() *ConnectionStri
...
+func (kcsb *ConnectionStringBuilder) AttachPolicyClientOptions(options *azcore.ClientOptions) *ConnectionStringBuilder {
 	requireNonEmpty(dataSource, kcsb.DataSource)
 	if options == nil {
 		kcsb.ClientOptions = options
 	}
 	return kcsb
 }
 
+// Create Kusto Conntection String that will be used for no auth mode. No auth token will be used for making requests.
+func (kcsb *ConnectionStringBuilder) WithNoAtuhentication() *ConnectionStringBuilder {
```
yogilad, line 287: No need for this API.
NoAuth mode should be the default behavior if no API is called.
DefaultAzureCredential should be an explicit choice, i.e. there should be an API named WithDefaultAzureCredential()

kusto/kcsb.go
```diff
@@ -423,6 +431,28 @@ func (kcsb *ConnectionStringBuilder) getTokenProvider() (*TokenProvider, error)
 
 }
 
+func (kcsb *ConnectionStringBuilder) initialiseWithCloudInfo(ci CloudInfo) {
```
yogilad, line 434: this feels wrong and too broad.
CloudInfo should not autoComplete KCSB, it should only affect the construction orf the specific credential and since those don't share all below decisions it's error prone.

Why not simply fill in the case-specific values in the token provider configuration object per each provider created
?

yogilad (approved):
Approved.
Prior to completion, please make sure to 
1) Test all auth flows (if you haven't)
2) Resolve all handled or irrelevant comments in the conversation section

kusto/cloudinfo.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return CloudInfo{}, err
+	}
+
+	// Handle internal server error as a special case and return as an error (to be consistent with other SDK's)
+	if resp.StatusCode >= http.StatusInternalServerError {
+		return CloudInfo{}, fmt.Errorf("error %s when querying endpoint %s", resp.Status, u.String())
+	}
+
+	//  return fmt.Errorf("error %s when querying endpoint %s", metadataResponse.StatusCode, fullMetadataEndpoint)
```
AsafMah, line 85: Remove commented code

kusto/cloudinfo.go
```diff
@@ -0,0 +1,115 @@
...
+	KustoServiceResourceID: defaultKustoServiceResourceId,
+	FirstPartyAuthorityURL: defaultFirstPartyAuthorityUrl,
+}
+
+// cache to query it once per instance
+var cloudInfoCache sync.Map
+
+func GetMetadata(ctx context.Context, kustoUri string) (CloudInfo, error) {
+	// retrieve &return if exists
+	clusterCachedCloudInfo, ok := cloudInfoCache.Load(kustoUri)
```
AsafMah, line 56: Doesn't this mean this if two clients will get here at the same time, both will do all of the work?

kusto/cloudinfo_test.go
```diff
@@ -0,0 +1,151 @@
...
+				KustoClientAppID:       "db662dc1-0cfe-4e1c-a843-19a68e65be58",
+				KustoClientRedirectURI: "https://microsoft/kustoclient",
+				KustoServiceResourceID: "https://kusto.dev.kusto.windows.net",
+				FirstPartyAuthorityURL: "https://login.microsoftonline.com/f8cdef31-a31e-4b4a-93e4-5f571e91255a",
+			},
+		},
+	}
+
+	ctx := context.Background()
+	for _, test := range tests {
```
AsafMah, line 133: Use modern style testing:
- t.Run to separate them
- Run in parallel
- Use assert.* instead of t.Errorf

kusto/kcsb.go
```diff
@@ -0,0 +1,485 @@
...
+
+// Creates a Kusto Connection string builder that will use existing authenticated az cli profile password.
+func (kcsb *ConnectionStringBuilder) WithAzCli() *ConnectionStringBuilder {
+	requireNonEmpty(dataSource, kcsb.DataSource)
+	kcsb.resetConnectionString()
+	kcsb.AzCli = true
+	return kcsb
+}
+
+/*
```
AsafMah, line 238: I think docstrings only work with // comments

kusto/kcsb.go
```diff
@@ -0,0 +1,485 @@
...
+			}
+			inOps := &azidentity.InteractiveBrowserCredentialOptions{}
+			inOps.ClientID = fetchedCI.KustoClientAppID
+			inOps.TenantID = kcsb.AuthorityId
+			inOps.RedirectURL = fetchedCI.KustoClientRedirectURI
+			inOps.ClientOptions = *cliOpts
+
+			cred, err := azidentity.NewInteractiveBrowserCredential(inOps)
+			if err != nil {
+				return nil, fmt.Errorf("Error : Couldn't retrieve client credentiels using Interactive Login. Error: %s", err)
```
AsafMah, line 331: Errors should be lowercase in go

kusto/kcsb.go
```diff
@@ -0,0 +1,485 @@
...
+	kcsb.DefaultAuth = true
+	return kcsb
+}
+
+// Method to be used for generating TokenCredential
+func (kcsb *ConnectionStringBuilder) newTokenProvider() (*TokenProvider, error) {
+	tkp := &TokenProvider{}
+	tkp.tokenScheme = BEARER_TYPE
+	// Fetches cloud meta data
+	fetchedCI, cierr := GetMetadata(context.Background(), kcsb.DataSource)
```
AsafMah, line 302: Isn't this exactly what we wanted not to do?
This does the web request to cloud info when the client is created.
It should be done only when the token is requested.

AsafMah (changes requested):

kusto/kcsb.go
asaharn: @AsafMah Only the first call for the new data source will be a web call. After that all will be cache access. 
the cloudSettingsInfo is required here to get tokenCredentials (defaulted with cloudSettinginfo)

AsafMah: 1. No, it's one call *per cluster*. Even if not, creating an object shouldn't do any long running actions.
Look what we do in all of the other SDKs.
2. Then lazily create them when needed, like in all other sdks.

kusto/cloudinfo.go
```diff
@@ -0,0 +1,116 @@
...
+	metadataClient := http.Client{Timeout: time.Duration(5) * time.Second}
+	req, err := http.NewRequest("GET", u.String(), nil)
+
+	if err != nil {
+		return CloudInfo{}, err
+	}
+	clusterCachedCloudInfo, ok = cloudInfoCache.Load(kustoUri)
+	if ok {
+		return clusterCachedCloudInfo.(CloudInfo), nil
+	}
```
AsafMah, line 76: How does this help? You're checking again after you have done the request, which is the heavy part.

kusto/cloudinfo.go
```diff
@@ -0,0 +1,116 @@
...
+	KustoClientAppID:       defaultKustoClientAppId,
+	KustoClientRedirectURI: defaultRedirectUri,
+	KustoServiceResourceID: defaultKustoServiceResourceId,
+	FirstPartyAuthorityURL: defaultFirstPartyAuthorityUrl,
+}
+
+// cache to query it once per instance
+var cloudInfoCache sync.Map
+
+func GetMetadata(ctx context.Context, kustoUri string) (CloudInfo, error) {
```
AsafMah, line 54: ctx argument is never used

kusto/cloudinfo_test.go
```diff
@@ -0,0 +1,152 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			s.code = test.code
+			s.payload = []byte(test.payload)
+			res, err := GetMetadata(ctx, s.urlStr()+"/"+test.name) // Adding test name to the path make sure multiple URL's can be cached
+			switch {
+			case err != nil && !test.err:
+				assert.FailNow(t, fmt.Sprintf("TestGetMetadata(%s): got err == %s, want err == nil", test.desc, err))
+			case err == nil && test.err:
+				assert.FailNow(t, fmt.Sprintf("TestGetMetadata(%s): got err == nil, want err != nil", test.desc))
+			}
```
AsafMah, line 144: Add assert.NoError before lline 149 and remove this switch.

kusto/kusto.go
```diff
@@ -26,85 +22,10 @@
 	queryToJson(ctx context.Context, db string, query Stmt, options *queryOptions) (string, error)
 }
 
-// Authorization provides the ADAL authorizer needed to access the resource. You can set Authorizer or
-// Config, but not both.
+// Authorization provides the Tokenprovider needed to acquire the auth token.
```
AsafMah, line 25: nit: TokenProvider

kusto/tokenprovider.go
```diff
@@ -0,0 +1,59 @@
...
+	err         error                  //To monitor errors in the process of init tokenprovider
+}
+
+// tokenProvider need to be received as reference, to reflect updations to the structs
+func (tkp *TokenProvider) AcquireToken(ctx context.Context) (string, string, error) {
+	if !isEmpty(tkp.customToken) {
+		return tkp.customToken, tkp.tokenScheme, nil
+	}
+
+	if tkp.tokenCred == nil {
```
AsafMah, line 30: Shouldn't the check be on "initOnce"?

kusto/tokenprovider.go
```diff
@@ -0,0 +1,59 @@
...
+			return "", "", tkp.err
+		}
+		// Update resource URI if MFA enabled
+		resourceURI := tkp.ci.KustoServiceResourceID
+		if tkp.ci.LoginMfaRequired {
+			resourceURI = strings.Replace(resourceURI, ".kusto.", ".kustomfa.", 1)
+		}
+		scopes := []string{fmt.Sprintf("%s/.default", resourceURI)}
+		tkp.scopes = scopes
+
```
AsafMah, line 43: You should wrap all of this in the init once

AsafMah (changes requested):
Good job with the lazy init, some comments.

kusto/cloudinfo.go
asaharn: The intention was to check, if the map is updated with the latest value before making an http request. and I guess the request is yet to be make here, since we will do client.Do after above?

kusto/tokenprovider.go
asaharn: @AsafMah  The sync.Once should not be used as a communication to check if it is already done the task. An initializer flag can be used to do that, but that we avoided by checking the tokenCred object itself.  Are we missing any cases if we use tokenCred as a check?

kusto/tokenprovider.go
AsafMah: Resolving this, see other note.

kusto/tokenprovider.go
```diff
@@ -0,0 +1,50 @@
...
+// tokenProvider need to be received as reference, to reflect updations to the structs
+func (tkp *TokenProvider) AcquireToken(ctx context.Context) (string, string, error) {
+	if !isEmpty(tkp.customToken) {
+		return tkp.customToken, tkp.tokenScheme, nil
+	}
+
+	if tkp.tokenCred == nil {
+		// initialise the tokenCredential and cloud info
+		tkp.initOnce.Do(tkp.init)
```
AsafMah, line 31: I didn't mean for you to copy-paste it to every provider, in fact in any case where you have to copy paste ask yourself if there's a better way.

Consider:

```go
tkp.initOnce.Do(func() {
tkp.init()

// Update resource URI if MFA enabled
resourceURI := tkp.ci.KustoServiceResourceID
if tkp.ci.LoginMfaRequired {
resourceURI = strings.Replace(resourceURI, ".kusto.", ".kustomfa.", 1)
}
scopes := []string{fmt.Sprintf("%s/.default", resourceURI)}
tkp.scopes = scopes
})
```

AsafMah (approved):

Issue #116, opened 2022-09-20
.show database Samples schema returns err "column of type System.String, which was unknown"


jesseward:
Running the `Mgmt(..)` query `.show database Samples schema` returns the error `DataTable had column of type System.String, which was unknown`. 

I am seeing this error on a number of databases I have been developing against and tests against the Samples demonstration database also return this error.

The following code can reproduce.

```go
package main

import (
	"context"
	"fmt"
	"io"
	"os"

	"github.com/Azure/azure-kusto-go/kusto"
	"github.com/Azure/go-autorest/autorest/azure/auth"
)

const (
	showTablesKql    = ".show tables | project TableName" // good query..
	showDbKql        = ".show database Samples schema"    // bad query
	dbName           = "Samples"                          // via https://help.kusto.windows.net
	connectionString = "https://help.kusto.windows.net"
)

func check(err error, where string) {
	if err != nil {
		fmt.Fprintf(os.Stderr, "ERROR: %s -> %v\n", where, err)
		os.Exit(1)
	}
}

func main() {
	azCreds, err := auth.NewAuthorizerFromCLIWithResource(connectionString)
	check(err, "authorizer")

	authorizer := kusto.Authorization{Authorizer: azCreds}
	client, err := kusto.New(connectionString, authorizer)
	check(err, "client")

	rows, err := client.Mgmt(context.Background(), dbName, kusto.NewStmt(showDbKql))
	check(err, "mgmt")

	for {
		row, err := rows.Next()
		if err != nil {
			if err == io.EOF {
				check(err, "eof")
			}
			check(err, "row")
		}
		fmt.Printf("%#v\n", row)
	}
}

```

Running the above returns
```sh
go run main.go 
ERROR: row -> DataTable had column of type System.String, which was unknown
exit status 1
```

Note that if you swap to another working query (eg `showTablesKql`), it confirms parsing of other mgmt commands are successful.

element-of-surprise:
At a quick glance, I'd guess something is causing this line to occur:
***@***.***/-/blob/kusto/internal/frames/v1/v1.go?L41


Which is in turn skipping the V1 frames type translation from happening on line:
https://sourcegraph.com/github.com/Azure/azure-kusto-go/-/blob/kusto/internal/frames/v1/v1.go?L47

V1 and V2 frames have different names for types and we have an internal type in Go we use to represent all column types.  Normally, what should happen, is when we receive a V1 frame and we translate it to the correct type. In this case:
[...]

PR #117, opened 2022-09-22
Fix Column.Type assignment. Was using string, now using types. Column value

jesseward:
#### Pull Request Description

This fixes the `v1` frame decoder bug that was resulting in the error `DataTable had column of type System.String, which was unknown`.

Fixes Azure/azure-kusto-go#116

---

#### Future Release Comment
Fixes `Mgmt(..)` query bug. Failure to unmarshall, caused by incorrect type assignment in v1 frame data.

**Fixes:**
- [Issue 116](https://github.com/Azure/azure-kusto-go/issues/116)

### DIFF
--- a/kusto/internal/frames/v1/v1.go
+++ b/kusto/internal/frames/v1/v1.go
@@ -35,22 +35,24 @@ type DataType struct {
 	DataType   string
 }
 
+func (dt DataType) rawType() string {
+	if dt.ColumnType != "" {
+		return dt.ColumnType
+	}
+	return dt.DataType
+}
+
 func (dt DataType) toColumn() (table.Column, error) {
 	col := table.Column{Name: dt.ColumnName}
 
-	if dt.ColumnType != "" {
-		var ok bool
-		col.Type, ok = translate[strings.ToLower(dt.ColumnType)]
-		if !ok {
-			return col, errors.ES(errors.OpMgmt, errors.KInternal, "DataTable.Columns(v1) had string entry with .ColumnType set to %s, which is not supported", dt.ColumnType)
-		}
-		return col, nil
+	if dt.rawType() == "" {
+		return col, errors.ES(errors.OpMgmt, errors.KInternal, "DataTable.Columns(v1) did not have ColumnType or DataType provided")
 	}
 
 	var ok bool
-	col.Type, ok = translate[strings.ToLower(dt.DataType)]
+	col.Type, ok = translate[strings.ToLower(dt.rawType())]
 	if !ok {
-		return col, errors.ES(errors.OpMgmt, errors.KInternal, "DataTable.Columns(v1) had entry with .DataType set to %q type, which is not supported", dt.DataType)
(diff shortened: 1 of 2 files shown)


kusto/internal/frames/v1/v1.go
```diff
@@ -39,7 +39,11 @@ func (dt DataType) toColumn() (table.Column, error) {
 	col := table.Column{Name: dt.ColumnName}
 
 	if dt.ColumnType != "" {
-		col.Type = types.Column(dt.ColumnType)
+		var ok bool
```
element-of-surprise, line 42: I think I want to make a cleanup to this and cover an extra error condition we aren't covering.  

You can find my version at:
https://go.dev/play/p/eEez1fBQ9M-

I also am including a test there that can go in the test file.  I haven't run this, but if it is off it should be barely off.

Again, thanks for doing this Jesse.

element-of-surprise (changes requested):
Thanks Jesse for this.  I won't be doing the final review, as I am no longer leading this effort and only tangentially involved.

As you'll see in my comment, I've made a few updates and added a test. you can include.  I think this will give us better coverage.

Cheers and thanks again.

element-of-surprise:
I added AsafMah to do the final review once the review comments are done.

kusto/internal/frames/v1/v1.go
element-of-surprise: Figured I'd ping on this issue, see if you still want to make the changes and push over.

jesseward: hey @element-of-surprise, I have now pushed the updated commit (with changes + unit test) into this branch

element-of-surprise (approved):

element-of-surprise:
@AsafMah : There are two tests failing here, but neither seem to be linked to these changes (bad tests it looks to me).

1st one:
=== CONT  TestError
    etoe_test.go:1500: 
        	Error Trace:	etoe_test.go:1500
        	Error:      	Received unexpected error:
        	            	Op(OpServConn): Kind(KClientArgs): cannot leave all Authoriztion fields as zero values
        	Test:       	TestError
--- FAIL: TestError (0.00s)

Looks like a test that skip wasn't turned on, since it is running without the setup environment.

2nd one:
=== CONT  TestNonProgressive/Cancelled_context
    statemachine_test.go:61: 
        	Error Trace:	statemachine_test.go:61
        	            				statemachine_test.go:521
        	            				statemachine_test.go:119
        	            				statemachine_test.go:508
        	Error:      	Not equal: 
        	            	expected: *errors.errorString(&errors.errorString{s:"context canceled"})
        	            	actual  : *errors.Error(&errors.Error{Op:0x0, Kind:0x2, Err:(*errors.errorString)(0xc00037a180), restErrMsg:[]uint8(nil), decoded:map[string]interface {}(nil), permanent:false, inner:(*errors.Error)(nil)})
        	Test:       	TestNonProgressive/Cancelled_context
        	Messages:   	wantErr: context canceled, gotErr: Kind(KInternal): non-progressive stream did not have DataSetCompletion frame

This looks like a test error or maybe even an error in the non-progressive stream.  But again, not related to this change.

element-of-surprise:
ping @AsafMah

PR #126, opened 2022-11-08
Remove locks where not needed

AsafMah:
Previously, the query client was locked for every request.

This created a long queue in some workflows.

Turns out this was unneeded by the code.

### DIFF
--- a/kusto/kusto.go
+++ b/kusto/kusto.go
@@ -195,9 +195,6 @@ const (
 // Note that the server has a timeout of 4 minutes for a query by default unless the context deadline is set. Queries can
 // take a maximum of 1 hour.
 func (c *Client) Query(ctx context.Context, db string, query Stmt, options ...QueryOption) (*RowIterator, error) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
-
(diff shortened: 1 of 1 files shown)


ag-ramachandran (approved):

kusto/kusto.go
```diff
@@ -335,7 +326,7 @@ func (c *Client) Mgmt(ctx context.Context, db string, query Stmt, options ...Mgm
 	return iter, nil
 }
 
-func (c *Client) setQueryOptions(ctx context.Context, op errors.Op, query Stmt, options ...QueryOption) (*queryOptions, error) {
+func setQueryOptions(ctx context.Context, op errors.Op, query Stmt, options ...QueryOption) (*queryOptions, error) {
```
yogilad, line 329: Isn't this a breaking change? 
(usage of the function changes from instance.Invoke to modul.Invoke)

yogilad:
@AsafMah , since mu is now used for one scenario, please name it accordingly (newIngestConnMu or something)
It was too widely used to begin with anyway

kusto/kusto.go
AsafMah: It's a private method (lowercase start), so we can change it at will.

PR #133, opened 2022-11-14
SafeStmt rework

ronmonetaMicro:
#### Pull Request Description

---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None


kusto/query_builder.go
```diff
@@ -506,6 +507,143 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+
+func (s Stmt) AddColumn(query string) Stmt {
+	return s.SafeAddName(query)
+}
+
+func (s Stmt) AddFunction(query string) Stmt {
+	return s.SafeAddName(query)
+}
+
+func (s Stmt) SafeAddName(query string) Stmt {
```
yogilad, line 526: Use the same name in C# (it would help making the link back to that logic)

kusto/query_builder.go
```diff
@@ -506,6 +507,143 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+	if !(strings.Contains(query, "'")) {
```
yogilad, line 527: You reordered the logic in a way which I can't say is correct.
Also, please do support this C# code section 

```
if (!forceNormalization
                && IsIdentifier(name)
```
The ask to exclude was only for 
```
&& !IsKeyword(name) 
```

kusto/query_builder.go
```diff
@@ -506,6 +507,143 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+	if !(query == "" || strings.HasPrefix(query, "[")) {
+		query = "[\"" + query + "\"]"
+	}
+
+	s.queryStr = s.queryStr + query
+	return s
+}
+
+func (s Stmt) AddQuotedString(value string, hidden bool) Stmt {
+	if value == "" {
```
yogilad, line 540: This was a null check.
I believe for an empty string the correct output should be a quoted empty string i.e.
```
""
```

kusto/query_builder.go
```diff
@@ -506,6 +507,143 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+		return s
+	}
+
+	literal := ""
```
yogilad, line 544: https://www.educative.io/answers/how-to-use-string-builder-in-golang

yogilad:
One more thing.
There's an example go file with statement examples.
Worth adding it there.

kusto/query_builder.go
```diff
@@ -506,6 +507,143 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
 	return s
 }
 
+func (s Stmt) AddDatabase(query string) Stmt {
```
yogilad, line 510: All public APIs should have a doc string starting with the API name

kusto/query_builder.go
ronmonetaMicro: If i understood correctly, this is the way to do a null check on string in Go (cannot directly check if it's nil). should we keep it like this then? im returning s because the actual effect of it would be no change to the statement that called this function. (and we are "returning" a stmt, so i cant return  an empty string)

kusto/query_builder.go
ronmonetaMicro: Didn't you say in the mail not to use a string builder? thats why i avoided it

yogilad: Sorry for the confusion. I was not aware of Go having a string builder class and meant this was not a requirement.
Since it exists, please use it :)

kusto/query_builder.go
```diff
@@ -507,35 +507,68 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
 	return s
 }
 
+// AddDatabase will add a normalized Database string to the Stmt. This is similar to the + operator on two strings,
```
yogilad, line 510: // AddDatabase will add a normalized Database string to the Statement..
// Given a variable with a value 'MyDatabase', AddDatabase(arg) will produce the following string '["MyDatabase"]'

For all below

yogilad (commented):
Seems ok.
Fix the last few comments and we'll close this

kusto/query_builder.go
```diff
@@ -506,6 +507,185 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+	for _, c := range query {
+		if !(((unicode.IsLetter(c) || unicode.IsDigit(c)) && unicode.In(c, unicode.ASCII_Hex_Digit)) || c == '_') {
+			return false
+		}
+	}
+	return true
+}
+
+// AddQuotedString escapes a string to be safely added to a stmt
+func (s Stmt) AddQuotedString(value string, hidden bool) Stmt {
```
AsafMah, line 572: You don't use this anywhere in the code. Shouldn't you quote the name in 'NormalizeName'?

kusto/query_builder.go
```diff
@@ -506,6 +507,185 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+}
+
+// AddFunction will add a normalized Function string to the Stmt. This is similar to the + operator on two strings,
+// except it only can be done with string constants. This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddFunction(query string) Stmt {
+	return s.NormalizeName(query, false)
+}
+
+// NormalizeName normalizes a string in order to be used safely in the engine.
+func (s Stmt) NormalizeName(query string, forceNormalization bool) Stmt {
```
AsafMah, line 535: Tests? for everything? we don't want to mess it up

AsafMah (changes requested):

kusto/query_builder.go
ronmonetaMicro: not accurate as this describe NormalizeName and not really these functions. i can maybe write 
// AddDatabase  - Given a variable with a value 'MyDatabase', AddDatabase(arg) will produce the following string '["MyDatabase"]' and will add it to a given stmt

kusto/query_builder.go
ronmonetaMicro: @yogilad So should we close it or should we write tests? (and if so - how? never done that :) )

kusto/query_builder.go
yogilad: "Given ... will append the following string to Stmt

kusto/query_builder.go
yogilad: A) PRs should include tests for new / changed code where possible.
B) here you have two valid options (and you'd probably want to cover both)
* Add unit tests which compose a query statement and validate the output string is as expected
* Add an end to end test which uses this syntax in queries and commands to be sure the output is accepted by the service (this can be an addition of scenarios to the existing query / command tests)

kusto/query_builder.go
yogilad: @AsafMah , the logic is taken from C#.
I agree NormalizeName() smells a bit fishy and has attack paths, but I would like to cover them with Ziv before making changes both here and in C#

kusto/query_builder.go
```diff
@@ -506,6 +507,185 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+func ShouldBeEscaped(c int32) bool {
+	if c <= unicode.MaxLatin1 {
+		return unicode.IsControl(c)
+	}
+	return true
+}
+
+// AddInt will add an int as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddInt(query int) Stmt {
+	s.queryStr = s.queryStr + fmt.Sprintf("%v", query)
```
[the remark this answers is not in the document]
ronmonetaMicro: nope, wrong copying. will be fixed :)

kusto/query_builder.go
```diff
@@ -506,6 +507,185 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+	} else {
+		query = "[\"" + query + "\"]"
+	}
+
+	s.queryStr = s.queryStr + query
+	return s
+}
+
+// IsIdentifier checks whether a given string is an identifier
+func IsIdentifier(query string) bool {
```
yogilad, line 556: Rename to requiresQuotation
Per our discussion, this returns false if the name is composed of only letters digits and underscores or if empty

kusto/query_builder.go
```diff
@@ -506,6 +507,185 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+func (s Stmt) AddFunction(query string) Stmt {
+	return s.NormalizeName(query, false)
+}
+
+// NormalizeName normalizes a string in order to be used safely in the engine.
+func (s Stmt) NormalizeName(query string, forceNormalization bool) Stmt {
+	if query == "" {
+		return s
+	}
+	if !forceNormalization && IsIdentifier(query) {
```
yogilad, line 539: if !force and !requireQoutes

kusto/query_builder.go
```diff
@@ -506,6 +507,185 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+		return s
+	}
+	if strings.HasPrefix(query, "[") {
```
yogilad, line 542: Drop this code below and always pass through AddQuotedString then surround in []

kusto/query_builder.go
```diff
@@ -506,6 +507,173 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+
+// AddByte will add a byte as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddByte(query byte) Stmt {
+	s.queryStr = s.queryStr + fmt.Sprintf("%d", query)
+	return s
+}
+
+// AddRune will add a rune as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddRune(query rune) Stmt {
+	s.queryStr = s.queryStr + fmt.Sprintf("%c", query)
```
yogilad, line 673: If this is a unicode char it should be quoted, otherwise KQL will mistreat it.

kusto/query_builder.go
```diff
@@ -506,6 +507,173 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+}
+
+// AddBool will add a bool as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddBool(query bool) Stmt {
+	s.queryStr = s.queryStr + fmt.Sprintf("%t", query)
+	return s
+}
+
+// AddByte will add a byte as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddByte(query byte) Stmt {
```
yogilad, line 666: Is byte a char or a number?
@AsafMah , what would you think should be the expected result?
Maybe we need two AddByte APIs (asChar AsInt)

AsafMah: It does %d so it adds it as a number.
In go "byte" is a number and "rune" is char.

kusto/query_builder.go
AsafMah: Rune is all chars, but yeah we should still escape it with the same logic as string.

yogilad: And you probably need to go through AddQuotedString to make sure the char is properly escaped.

yogilad: On third thought, I don't see a usable scenario for this API?

kusto/query_builder.go
yogilad: Per the below comment on Rune, maybe simply keep this as a number.

kusto/query_builder.go
ronmonetaMicro: So should we keep them? both rune and byte?

kusto/query_builder.go
yogilad: per our discussion let's leave this API as is

kusto/query_builder.go
yogilad: per our discussion, let's remove this API completely

kusto/query_builder_test.go
```diff
@@ -598,3 +598,319 @@ func buildQueryStr(query string, params Definitions) string {
 	}
 	return query
 }
+
+func TestNormalizeName(t *testing.T) {
```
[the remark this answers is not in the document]
AsafMah: For free:
```
		{
			desc: "Success string with numbers",
			query: "test123",
			forceNormalization: false,
			want: NewStmt("test123"),
		},
		{
			desc: "Success string that starts with a number",
			query: "123test",
			forceNormalization: false,
			want: NewStmt("[\"123test\"]"),
		},
		{
			desc: "Success string that starts with an underscore",
			query: "_test",
			forceNormalization: false,
			want: NewStmt("_test"),
		},
		{
			desc: "Success string that has an underscore",
			query: "test_123",
			forceNormalization: false,
			want: NewStmt("test_123"),
		},
```

kusto/query_builder.go
```diff
@@ -506,6 +507,171 @@
...
+}
+
+// AddFloat64 will add a Float64 as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddFloat64(query float64) Stmt {
+	s.queryStr = s.queryStr + fmt.Sprintf("%f", query)
+	return s
+}
+
+// AddComplex64 will add a Complex64 as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddComplex64(query complex64) Stmt {
```
AsafMah, line 652: I think you should remove the complexes, kusto does not have an equivalent type, it's just confusing.

kusto/query_builder.go
```diff
@@ -506,6 +507,171 @@
...
+	s.queryStr = s.queryStr + fmt.Sprintf("%t", query)
+	return s
+}
+
+// AddByte will add a byte as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddByte(query byte) Stmt {
+	s.queryStr = s.queryStr + fmt.Sprintf("%d", query)
+	return s
+}
+
```
AsafMah, line 674: Why did we remove the AddRune method? it should just be treated as a one-length string

kusto/query_builder.go
```diff
@@ -506,6 +507,171 @@
...
+// AddInt will add an int as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddInt(query int) Stmt {
+	s.queryStr = s.queryStr + fmt.Sprintf("%d", query)
+	return s
+}
+
+// AddFloat32 will add a Float32 as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddFloat32(query float32) Stmt {
```
AsafMah, line 640: All of these AddX methods are kind of unsafe

Consider someone passing NaN here, kusto won't be able to parse it.

I think the logic to these should be the same as in "validate" - I.E calling the type constructors.
That would make it "safer", and you can add other kusto types like Date and such, and unify the logic between both of them.

If you want we can discuss it with yochai.

kusto/query_builder.go
```diff
@@ -506,6 +507,171 @@
...
+			literal.WriteString("\\t")
+
+		case '\v':
+			literal.WriteString("\\v")
+
+		default:
+			if !ShouldBeEscaped(c) {
+				literal.WriteString(string(c))
+			} else {
+				literal.WriteString(fmt.Sprintf("\\u%x", c))
```
AsafMah, line 615: This is wrong, the hex should always be in length four. See what your unit test result does in kusto. 
change to `\\u%04x`

AsafMah (changes requested):

kusto/query_builder.go
ronmonetaMicro: as discussed with @yogilad , due to the high similarity in function to strings, we saw no real need for this method. indeed, the cost of adding it back if desired is minimal, so shouldnt be a problem if we still want to

element-of-surprise:
So, I think this PR is going to break security constraints for the Stmt type, if I'm reading the changes correctly.

There is a bunch of methods that have (query string) as arguments.  

When I built this, I added a security constraint to Stmt so that you cannot add adhoc strings to a Stmt, you can only add Go string constants.  This is a preventative  measure to prevent SQL injection like attacks where the user just passes something from some unsanitized input  directly to Kusto.

To be able to add non string constants, you had to use the "unsafe" library.  This gave an extra level of protection, requiring you to say what you are doing is unsafe and read why it is unsafe.  

Maybe NormalizeName() and various KQL features are letting you completely get around this problem?  Just pointing out the purpose of `stringConstant` and the `unsafe` package.

kusto/query_builder.go
```diff
@@ -64,147 +65,146 @@ type ParamType struct {
...
+		default:
+			return false
 		}
-		return nil
 	case types.Long:
-		if _, ok := p.Default.(int64); !ok {
-			return fmt.Errorf("the .Type was %s, but the value was a %T", p.Type, p.Default)
-		}
-		return nil
+		_, ok := val.(int64)
```
AsafMah, line 96: Aren't int & int32 also valid to convert to long?

kusto/query_builder.go
```diff
@@ -64,147 +65,146 @@ type ParamType struct {
...
+		return ok
 	case types.Real:
-		if _, ok := p.Default.(float64); !ok {
-			return fmt.Errorf("the .Type was %s, but the value was a %T", p.Type, p.Default)
-		}
-		return nil
+		_, ok := val.(float64)
```
AsafMah, line 99: float32 too?

kusto/query_builder.go
```diff
@@ -506,6 +435,166 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+	return true
+}
+
+// AddInt will add an int as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddInt(query int) Stmt {
+	return s.addBase(types.Int, query)
+}
+
+// AddFloat32 will add a Float32 as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddFloat32(query float32) Stmt {
```
AsafMah, line 567: This (and 64) is wrong.

The type should be types.Float.

kusto/query_builder.go
```diff
@@ -506,6 +435,166 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+// ShouldBeEscaped Checks whether a rune should be escaped or not based on it's type.
+func ShouldBeEscaped(c int32) bool {
+	if c <= unicode.MaxLatin1 {
+		return unicode.IsControl(c)
+	}
+	return true
+}
+
+// AddInt will add an int as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddInt(query int) Stmt {
```
AsafMah, line 562: AddInt32,and AddInt64 (for long)

kusto/query_builder.go
```diff
@@ -506,6 +435,166 @@ func NewStmt(query stringConstant, options ...StmtOption) Stmt {
...
+}
+
+// AddFloat32 will add a Float32 as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddFloat32(query float32) Stmt {
+	return s.addBase(types.Decimal, query)
+}
+
+// AddFloat64 will add a Float64 as a string to the Stmt.  This allows dynamically building of a query from a root Stmt.
+func (s Stmt) AddFloat64(query float64) Stmt {
+	return s.addBase(types.Decimal, query)
```
AsafMah, line 573: Also add AddDecimal? I guess it will have to accept interface{} or use generics

kusto/query_builder.go
```diff
@@ -64,147 +65,146 @@ type ParamType struct {
...
+		_, ok := val.(bool)
+		return ok
 	case types.DateTime:
-		if _, ok := p.Default.(time.Time); !ok {
-			return fmt.Errorf("the .Type was %s, but the value was a %T", p.Type, p.Default)
-		}
-		return nil
+		_, ok := val.(time.Time)
+		return ok
 	case types.Dynamic:
```
AsafMah, line 81: This should return true, because any type could potentially be OK for dynamic. (that's why one of tests fail)

AsafMah (changes requested):

kusto/query_builder.go
ronmonetaMicro: based on what i've seen (and in the old tests as well) - it semms like Long is int64. am i missing something?

kusto/query_builder.go
ronmonetaMicro: As our discussion above - i think int64 is only for long. and we can also add a addint32 - but i'll admit i dont think it will be used, most likely addint will be the "goto" one. WDYT?

kusto/query_builder.go
ronmonetaMicro: you mean in the type passed into addbase? we've built the other function using types.Decimal - and there is also no types.Float 
image

Issue #138, opened 2022-11-25
API to set client version 


asaharn:
In current version of SDK a way to set client Version (similar to java sdk's [ConnectionStringBuilder.setClientVersionForTracing](https://github.com/Azure/azure-kusto-java/blob/master/data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java#L163)) is missing
API can be implemented to set the client version so that it can be used for tracing and for connector/application specific stats which could be helpful for tracking clients' usage.

AsafMah:
Added long ago :)

PR #142, opened 2022-11-29
* introduced API for setting client version name

asaharn:
#### Pull Request Description

Provided API ConnectionStringBuilder.SetClientVersionForTracing()

client Version Name will be appended to client version header. 

issue ref: #138 
---

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None

### DIFF
--- a/kusto/conn.go
+++ b/kusto/conn.go
@@ -39,10 +39,11 @@ type conn struct {
 	auth                           Authorization
 	endMgmt, endQuery, streamQuery *url.URL
 	client                         *http.Client
+	versionName                    string
 }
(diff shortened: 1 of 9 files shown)


AsafMah:
Mmm The structure of go makes this a bit annoying.

I think we should first generalize it into a struct, since we might want to add similar things later (like client request id and such).

Also we can share code between kusto/conn (for queries) and conn/conn (for streaming ingestion).

Also of course add a test.

AsafMah (changes requested):
see comment

PR #143, opened 2022-12-01
Fixed leaks

AsafMah:
- Http error now closes the body
- renewResources now ticks every 30 seconds to not keep the goroutine hanging
- Closed resources in tests

### DIFF
--- a/kusto/ingest/internal/resources/resources.go
+++ b/kusto/ingest/internal/resources/resources.go
@@ -174,11 +174,18 @@ func (m *Manager) Close() {
 }
 
 func (m *Manager) renewResources() {
-	tick := time.NewTicker(1 * time.Hour)
+	tickDuration := 30 * time.Second
+	tickTotal := 1 * time.Hour
+	tick := time.NewTicker(tickDuration)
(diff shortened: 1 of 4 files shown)


kusto/ingest/internal/resources/resources.go
```diff
@@ -174,11 +174,18 @@ func (m *Manager) Close() {
...
+	tickDuration := 30 * time.Second
+	tickTotal := 1 * time.Hour
+	tick := time.NewTicker(tickDuration)
+	count := 0 * time.Second
 	for {
 		select {
 		case <-tick.C:
-			m.fetchRetry(context.Background())
+			count += tickDuration
+			if count >= tickTotal {
```
yogilad, line 185: Making sure there's some code out there which - when trying to get the resources - tries to refresh it synchronously if the expected duration has expired by a factor of 2 or 3

yogilad (approved):
with suggestion

kusto/ingest/internal/resources/resources.go
AsafMah: Another PR

PR #144, opened 2022-12-01
Synchronously fetch resources

AsafMah:
Synchronously fetch resources if didn't update in 2 hours.


kusto/ingest/internal/resources/resources.go
```diff
@@ -361,6 +366,11 @@ func (m *Manager) fetchRetry(ctx context.Context) {
 // Resources returns information about the ingestion resources. This will used cached information instead
 // of fetching from source.
 func (m *Manager) Resources() (Ingestion, error) {
+	lastFetchTime, ok := m.lastFetchTime.Load().(time.Time)
+	if !ok || lastFetchTime.Add(2*fetchInterval).After(time.Now().UTC()) {
+		m.fetchRetry(context.Background())
```
yogilad, line 371: Fetch retry may hang forever.
You should (at least optionally and here explicitly) limit the retry count (3?) and return an error if failed

kusto/ingest/internal/resources/resources.go
```diff
@@ -175,14 +178,14 @@ func (m *Manager) Close() {
...
-	tickTotal := 1 * time.Hour
+
 	tick := time.NewTicker(tickDuration)
 	count := 0 * time.Second
 	for {
 		select {
 		case <-tick.C:
 			count += tickDuration
-			if count >= tickTotal {
+			if count >= fetchInterval {
```
yogilad, line 188: Consider if you want fetch retry to loop for ever if the service is down for example

yogilad (commented):
Also not the New() call to fetch will fail the resource manager and ingest client creation if the service is down

kusto/ingest/internal/resources/resources.go
AsafMah: Huh, fetchRetry has an "attempts" variable that we increase, but never use..

kusto/ingest/internal/resources/resources.go
AsafMah: Not sure I understand how this comment is different from the last one.

kusto/ingest/internal/resources/resources.go
AsafMah: Fixed.

kusto/ingest/internal/resources/resources.go
AsafMah: Fixed.

kusto/ingest/internal/resources/resources.go
```diff
@@ -152,6 +154,7 @@ func New(client mgmter) (*Manager, error) {
 	}
 
 	m.kustoTokenCacheExpiration = time.Now().UTC()
+	m.lastFetchTime.Store(time.Now().UTC())
```
yogilad, line 157: I'd drop this line.
Consider what will happen if the initial fetch will fail

kusto/ingest/internal/resources/resources.go
```diff
@@ -152,6 +154,7 @@ func New(client mgmter) (*Manager, error) {
...
 	go m.renewResources()
```
yogilad, line 158: This API eventually uses m.fetchLock.
it's best to move it past line 161

kusto/ingest/internal/resources/resources.go
```diff
@@ -350,17 +359,24 @@ func (m *Manager) fetchRetry(ctx context.Context) {
 		cancel()
 		if err != nil {
 			attempts++
-			//log.Printf("problem fetching the resources from Kusto Mgmt(attempt %d): %s", attempts, err)
 			time.Sleep(10 * time.Second)
```
yogilad, line 362: You sleep only to loop back into the exist condition, i.e. worst case scenario you'd sleep 4 times rather then 3

AsafMah: Fixed

kusto/ingest/internal/resources/resources.go
AsafMah: Line 161 does not acquire the lock, just initializes the lock object.
Either way - I removed it.

kusto/ingest/internal/resources/resources.go
AsafMah: I removed the fetch in the ctor entirely.
It's weird that it does an action that can take seconds when constructing an ingestor object..

Now the background loop will immediately start one, and if for some reason it doesn't it will fall back to synchronous (since we have a condition if lastFetchTime is not set)

kusto/ingest/internal/resources/resources.go
yogilad: I meant you run a go routine using the lock object **before** you initialize it.
This seems prone to race conditions.

kusto/ingest/internal/resources/resources.go
yogilad: Not really ;)

AsafMah: Oops! fixed

yogilad (approved):

PR #152, opened 2022-12-11
Removed forced check for core.windows.net

AsafMah:
The long term solution is endpoints validation, but this works for now.

### DIFF
--- a/kusto/ingest/internal/resources/resources.go
+++ b/kusto/ingest/internal/resources/resources.go
@@ -46,7 +46,7 @@ type URI struct {
 // parse parses a string representing a Kutso resource URI.
 func parse(uri string) (*URI, error) {
 	// Regex representing URI that is expected:
-	// https://(\w+).(queue|blob|table).core.windows.net/([\w,-]+)\?(.*)
+	// https://(\w+).(queue|blob|table).[\w.]+/([\w,-]+)\?(.*)
 
 	u, err := url.Parse(uri)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


kusto/ingest/internal/resources/resources.go
```diff
@@ -46,7 +46,7 @@ type URI struct {
 // parse parses a string representing a Kutso resource URI.
 func parse(uri string) (*URI, error) {
 	// Regex representing URI that is expected:
-	// https://(\w+).(queue|blob|table).core.windows.net/([\w,-]+)\?(.*)
+	// https://(\w+).(queue|blob|table).[\w.]+/([\w,-]+)\?(.*)
```
yogilad, line 49: This comment is no longer relevant

kusto/ingest/internal/resources/resources.go
```diff
@@ -57,10 +57,6 @@ func parse(uri string) (*URI, error) {
 		return nil, fmt.Errorf("URI scheme must be 'https', was '%s'", u.Scheme)
 	}
 
-	if !strings.HasSuffix(u.Hostname(), ".core.windows.net") {
-		return nil, fmt.Errorf("URI hostname does not end with '.core.windows.net'")
-	}
-
 	hostSplit := strings.Split(u.Hostname(), ".")
 	if len(hostSplit) != 5 {
```
yogilad, line 61: the assumption here is _resource.resource_type.core.[dns suffix]_
The condition should better be changed to >4 cause we can't assure the DNS suffix for the cloud would only have two components.

kusto/ingest/internal/resources/resources.go
```diff
@@ -57,10 +57,6 @@ func parse(uri string) (*URI, error) {
...
 		return nil, fmt.Errorf("URI(%s) is invalid: had incorrect URL path before '.core.windows.net'", uri)
```
yogilad, line 62: This comment is incorrect

```suggestion
		return nil, fmt.Errorf("Storage URI (%s) is invalid'", uri)
```

kusto/ingest/internal/resources/resources.go
AsafMah: Why? we still check the rest of the uri, I changed the regex to fit any domain.

kusto/ingest/internal/resources/resources.go
AsafMah: Fixed

yogilad (approved):

PR #157, opened 2022-12-25
Fix msi opts

### LINKED ISSUE #156: Miss init method when using MsiAuthentication way to generate TokenCredential
init method was missed when using MsiAuthentication way to generate TokenCredential in `./kusto/kcsb.go` file, it should be added like this:

```
case kcsb.MsiAuthentication:
    init = func(ci *CloudInfo, cliOpts *azcore.ClientOptions, appClientId string) (azcore.TokenCredential, error) {
	    opts := &azidentity.ManagedIdentityCredentialOptions{}
	    if kcsb.ClientOptions != nil {
		    opts.ClientOptions = *kcsb.ClientOptions
	    }
	    if !isEmpty(kcsb.ManagedServiceIdentity) {
		    opts.ID = azidentity.ClientID(kcsb.ManagedServiceIdentity)
	    }
    
	    cred, err := azidentity.NewManagedIdentityCredential(opts)
	    if err != nil {
		    return nil, fmt.Errorf("error: Couldn't retrieve client credentials using Managed Identity: %s", err)
	    }
	    tkp.tokenCred = cred
    }
```

if `init==nil`, it won't call `setInit()` and `tokenWrapper()`, 

and the scope is assigned a value in the tokenWrapper method: `scopes := []string{fmt.Sprintf("%s/.default", resourceURI)}`, 

without assigning a value to the scopes, you will end up with an error when calling the `GetToken` method in `managed_identity_credential.go`

```
func (c *ManagedIdentityCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (azcore.AccessToken, error) {
	if len(opts.Scopes) != 1 {
		err := errors.New(credNameManagedIdentity + ": GetToken() requires exactly one scope")
		return azcore.AccessToken{}, err
	}
	...
}
```

AsafMah:
Fixes #156 and corrects typos.

### DIFF
--- a/kusto/kcsb.go
+++ b/kusto/kcsb.go
@@ -290,13 +290,13 @@ func (kcsb *ConnectionStringBuilder) newTokenProvider() (*TokenProvider, error) 
 	switch {
 	case kcsb.InteractiveLogin:
 		init = func(ci *CloudInfo, cliOpts *azcore.ClientOptions, appClientId string) (azcore.TokenCredential, error) {
-			inOps := &azidentity.InteractiveBrowserCredentialOptions{}
-			inOps.ClientID = ci.KustoClientAppID
-			inOps.TenantID = kcsb.AuthorityId
-			inOps.RedirectURL = ci.KustoClientRedirectURI
-			inOps.ClientOptions = *cliOpts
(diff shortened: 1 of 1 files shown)


asaharn (approved):
Looks good to me.

PR #158, opened 2022-12-29
Added alignment of headers according to spec

AsafMah:
Added headers to kscb, tests, etc


kusto/conn.go
```diff
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/google/uuid"
```
yogilad, line 11: odd auto-align decision

kusto/conn.go
```diff
@@ -39,10 +37,11 @@ type conn struct {
 	auth                           Authorization
 	endMgmt, endQuery, streamQuery *url.URL
 	client                         *http.Client
+	application, user, version     string
```
yogilad, line 40:
```suggestion
	applicationForTracing, userForTracing, clientVersion     string
```

kusto/conn.go
```diff
@@ -218,6 +207,30 @@ func (c *conn) doRequest(ctx context.Context, execType int, db string, query Stm
 	return op, header, resp.Header, body, nil
 }
 
+func (c *conn) getHeaders(properties requestProperties) http.Header {
+	header := http.Header{}
+	header.Add("Accept", "application/json")
+	header.Add("Accept-Encoding", "gzip")
+	header.Add("Content-Type", "application/json; charset=utf-8")
+	header.Add("x-ms-version", "2019-02-13")
+	header.Add("x-ms-client-request-id", "KGC.execute;"+uuid.New().String())
```
yogilad, line 216: Don't we support user provided ClientRequestId in Go?

kusto/ingest/internal/conn/conn.go
```diff
@@ -32,17 +31,18 @@ var BuffPool = sync.Pool{
...
-	baseURL    *url.URL
-	reqHeaders http.Header
-	client     *http.Client
-	done       chan struct{}
+	auth                       kusto.Authorization
+	baseURL                    *url.URL
+	reqHeaders                 http.Header
+	client                     *http.Client
+	done                       chan struct{}
+	application, user, version string
```
yogilad, line 39: same as with Kusto/conn.go

kusto/ingest/internal/conn/conn.go
```diff
@@ -51,15 +51,18 @@ func New(endpoint string, auth kusto.Authorization, client *http.Client) (*Conn,
 		).SetNoRetry()
 	}
 
-	return newWithoutValidation(endpoint, auth, client)
+	return newWithoutValidation(endpoint, auth, client, application, user, version)
 }
 
-func newWithoutValidation(endpoint string, auth kusto.Authorization, client *http.Client) (*Conn, error) {
+func newWithoutValidation(endpoint string, auth kusto.Authorization, client *http.Client, application, user, version string) (*Conn, error) {
```
yogilad, line 57: Why is this a duplicate of Kusto/conn.go
Also, for streamingingestion you can pass CRP so this should look similar to the other conn.

kusto/ingest/streaming.go
```diff
@@ -32,7 +32,7 @@ var FileIsBlobErr = errors.ES(errors.OpIngestStream, errors.KClientArgs, "blobst
 // More information can be found here:
 // https://docs.microsoft.com/en-us/azure/kusto/management/create-ingestion-mapping-command
 func NewStreaming(client QueryClient, db, table string) (*Streaming, error) {
-	streamConn, err := conn.New(client.Endpoint(), client.Auth(), client.HttpClient())
+	streamConn, err := conn.New(client.Endpoint(), client.Auth(), client.HttpClient(), "", "", "")
```
yogilad, line 35: Is not having CRPs here mapped as a parity gap, or is it the same in all SDKs?

kusto/kcsb.go
```diff
@@ -28,9 +34,15 @@ type ConnectionStringBuilder struct {
 	RedirectURL                      string
 	DefaultAuth                      bool
 	ClientOptions                    *azcore.ClientOptions
+	ApplicationForTracing            string
+	UserForTracing                   string
+	versionForTracing                string
```
yogilad, line 39: This is not a KCSB capability

kusto/kcsb.go
```diff
@@ -163,6 +175,35 @@ func (kcsb *ConnectionStringBuilder) resetConnectionString() {
...
+func (kcsb *ConnectionStringBuilder) setDefaultTracingValues() {
+	kcsb.ApplicationForTracing = filepath.Base(os.Args[0])
+	kcsb.UserForTracing = getOsUser()
+	kcsb.versionForTracing = buildHeaderFormat(StringPair{Key: "Kusto.Go.Client", Value: version.Kusto}, StringPair{Key: "Go", Value: runtime.Version()})
+}
+
+func getOsUser() string {
+	var final string
+	current, err := user.Current()
+	if err != nil || current.Username == "" {
```
yogilad, line 191: nit: this is odd cause you used the checked value in the else
Reverse the condition perhaps?
You can also consider scope reduction (if {return} X N)

kusto/ingest/streaming.go
yogilad: Also, should we not take this from 'client'?

kusto/kusto.go
```diff
@@ -381,6 +386,27 @@ func (c *Client) HttpClient() *http.Client {
 	return c.http
 }
 
+func (c *Client) Application() string {
+	if conn, ok := c.conn.(*conn); ok {
```
yogilad, line 390: not sure I understand.

kusto/conn.go
AsafMah: Nope, but I'll add

kusto/kcsb.go
AsafMah: WDYM? I'll move the logic to another class, but we still need getters/setters for these in KCSB

kusto/ingest/internal/conn/conn.go
AsafMah: conn.go is a private type, so we need to duplicate it here in this package.
Alternatively we can have this functionality be a part of the regular "conn" under an interface and reduce the duplication maybe

kusto/ingest/streaming.go
AsafMah: No SDK lets you send CRP to streaming ingestion

AsafMah: In fact, the endpoint doesn't expect it either..

AsafMah:
@yogilad I refactored it to remove the logic out of KSCB, take another look.

kusto/client_details.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	if isEmpty(final) {
+		final = NONE
+	}
+
+	return final
+}
+
+var escapeRegex = regexp.MustCompile("[\\r\\n\\s{}|]+")
```
yogilad, line 66: What's the Go best practice on this?
Should also consts be grouped some place or is it ok to spread them around?

yogilad (approved):

kusto/client_details.go
AsafMah: I'll put it with the other consts on the top.

yogilad (approved):

PR #161, opened 2023-01-23
Add validation for trusted endpoints

ohadbitt:


kusto/trustedEndpoints/trustedEndpoints.go
```diff
@@ -0,0 +1,240 @@
+package truestedEndpoints
```
AsafMah, line 1: file folder names - snake_case, not camelCase

kusto/trustedEndpoints/wellKnownKustoEndpoints.go
```diff
@@ -0,0 +1,90 @@
+package truestedEndpoints
```
AsafMah, line 1: So we weren't on the same page about it.. have the normal json file here and use https://pkg.go.dev/embed

kusto/test/etoe/etoe_test.go
```diff
@@ -99,7 +100,7 @@ type mgmtFunc func(ctx context.Context, db string, query kusto.Stmt, options ...
 
 type queryJsonFunc func(ctx context.Context, db string, query kusto.Stmt, options ...kusto.QueryOption) (string, error)
 
-func TestQueries(t *testing.T) {
+func TestQueries(t *testing.T) { // running
```
AsafMah, line 103: remove those comments in this file?

go.mod
```diff
@@ -33,7 +33,9 @@ require (
 	github.com/kr/text v0.1.0 // indirect
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/samber/lo v1.37.0 // indirect
```
AsafMah, line 36: make it a direct dependency. `go mod add github.com/samber/lo; go mod tidy`

kusto/trustedEndpoints/trustedEndpoints.go
```diff
@@ -0,0 +1,240 @@
...
+
+type WellKnownKustoEndpointsDataStruct struct {
+	AllowedEndpointsByLogin map[string]AllowedEndpoints
+}
+
+func createInstance() *TrustedEndpoints {
+	matchers := map[string]*FastSuffixMatcher{}
+	wellKnownData := WellKnownKustoEndpointsDataStruct{}
+
+	json.Unmarshal([]byte(WellKnownKustoEndpointsJson), &wellKnownData)
```
AsafMah, line 31: handle error

kusto/trustedEndpoints/trustedEndpoints.go
```diff
@@ -0,0 +1,240 @@
...
+		return NewFastSuffixMatcher(rules)
+	}
+
+	if rules == nil || len(rules) == 0 {
+		return existing, nil
+	}
+
+	list := rules[:]
+
+	for _, elem := range values(existing.rules) {
```
AsafMah, line 150: why do you need "values"? why not simply `range existing.rules`?

kusto/trustedEndpoints/trustedEndpoints.go
```diff
@@ -0,0 +1,240 @@
...
+func CreateFastSuffixMatcherFromExisting(rules []MatchRule, existing *FastSuffixMatcher) (*FastSuffixMatcher, error) {
+	if existing == nil || len(existing.rules) == 0 {
+		return NewFastSuffixMatcher(rules)
+	}
+
+	if rules == nil || len(rules) == 0 {
+		return existing, nil
+	}
+
+	list := rules[:]
```
AsafMah, line 148: I'm pretty sure this doesn't create a copy in go, if that's what you intended

kusto/trustedEndpoints/trustedEndpoints.go
```diff
@@ -0,0 +1,240 @@
...
+	if replace {
+		trusted.additionalMatcher = nil
+	}
+
+	matcher, err := CreateFastSuffixMatcherFromExisting(rules, trusted.additionalMatcher)
+	trusted.additionalMatcher = matcher
+	return err
+}
+
+/**
```
AsafMah, line 174: This is not how go doc strings work

AsafMah (changes requested):

go.mod
ohadbitt: guess you meant "go get"
anyway tidy did it

kusto/trustedEndpoints/trustedEndpoints.go
ohadbitt: to flat the list

kusto/trustedEndpoints/trustedEndpoints.go
ohadbitt: it creates a slice from the whole array

kusto/trustedEndpoints/trustedEndpoints.go
AsafMah: It does not do it though, you just create an array over the values and iterate over it, it's the same as normally iterating over the map.

The flattening is done with the "elem..." 


I changed the code to this, and all of the tests still pass:

```go
func createFastSuffixMatcherFromExisting(rules []MatchRule, existing *FastSuffixMatcher) (*FastSuffixMatcher, error) {
	if existing == nil || len(existing.rules) == 0 {
		return newFastSuffixMatcher(rules)
	}

	if rules == nil || len(rules) == 0 {
		return existing, nil
	}

	for _, elem := range existing.rules {
		rules = append(rules, elem...)
	}

	return newFastSuffixMatcher(rules)
}
```

kusto/trustedEndpoints/trustedEndpoints.go
AsafMah: You don't need it then, see above

kusto/trustedEndpoints/trustedEndpoints.go
AsafMah: Close, but you need to add the name of the method to the start of the doc: 
`// AddTrustedHosts Add or set a list of trusted endpoints rules`
`func (trusted *TrustedEndpoints) AddTrustedHosts(rules []MatchRule, replace bool) error {`

I'm surprised whatever plugin you use doesn't  auto correct this

kusto/trustedEndpoints/trustedEndpoints.go
AsafMah: You didn't push? the folder name is still "trustedEndpoints" and not "trusted_endpoints"

AsafMah (changes requested):
go over the comments again

kusto/trustedEndpoints/trustedEndpoints.go
ohadbitt: yea could be nice

PR #167, opened 2023-03-01
fix: Support storage uri to support all clouds

AsafMah:
Fixes issue that storage accounts only supported blob.windows.net


kusto/ingest/internal/queued/queued.go
```diff
@@ -273,7 +273,8 @@ func (i *Ingestion) upstreamContainer() (*azblob.Client, string, error) {
 	}
 
 	storageURI := mgrResources.Containers[rand.Intn(len(mgrResources.Containers))]
-	serviceURL := fmt.Sprintf("https://%s.blob.core.windows.net?%s", storageURI.Account(), storageURI.SAS().Encode())
+	storageUrl := storageURI.URL()
+	serviceURL := fmt.Sprintf("%s://%s?%s", storageUrl.Scheme, storageUrl.Host, storageURI.SAS().Encode())
```
yogilad, line 277: Aren't you missing the resource itself?
{scheme}://{host}/{resource}?{SAS}

yogilad (approved):

PR #175, opened 2023-03-28
Add no redirects

AsafMah:
### Security
* No redirects are allowed by default.


kusto/test/etoe/etoe_test.go
```diff
@@ -1967,6 +1971,52 @@ func assertErrorsMatch(t *testing.T, got, want error) bool {
...
+
+			// Change field using reflection.
+			endQuery := reflect.
+				ValueOf(client).
+				Elem().
+				FieldByName("conn").
+				Elem().
+				Elem().
+				FieldByName("endQuery")
+			reflect.NewAt(endQuery.Type(), goUnsafe.Pointer(endQuery.UnsafeAddr())).Elem().Set(reflect.ValueOf(urll))
```
yogilad, line 2009: Why do you need this level of complexity?
Can't you return an KustoClientError with an explicit message saying Redirects Not Allowed?
Mind this is a requirement, or users will fail without knowing why.

kusto/cloudinfo.go
```diff
@@ -64,8 +65,10 @@ func GetMetadata(kustoUri string, httpClient *http.Client) (CloudInfo, error) {
 		if err != nil {
 			return CloudInfo{}, err
 		}
-
-		u.Path = metadataPath
+		if !strings.HasPrefix(u.Path, "/") {
+			u.Path = "/" + u.Path
+		}
+		u = u.JoinPath(metadataPath)
```
yogilad, line 71: Do you need to prepand '/' when using url.JoinPath?

yogilad (approved):

kusto/cloudinfo.go
AsafMah: For some reason, the "JoinPath" gives unexpected results without it.
Maybe I should add a comment?

yogilad: won't hurt

PR #176, opened 2023-04-04
Replace non-ascii characters in headers to be in line with the service.

AsafMah:
### Fixed
* Replace non-ascii characters in headers to be in line with the service.

### DIFF
--- a/kusto/conn.go
+++ b/kusto/conn.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"unicode"
 
 	"github.com/Azure/azure-kusto-go/kusto/data/errors"
 	"github.com/Azure/azure-kusto-go/kusto/internal/frames"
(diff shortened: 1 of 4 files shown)


kusto/conn.go
```diff
@@ -192,6 +193,21 @@ func (c *Conn) doRequestImpl(
 	headers http.Header,
 	errorContext string) (http.Header, io.ReadCloser, error) {
 
+	// Replace non-ascii chars in headers with '?'
+	for _, values := range headers {
+		var builder strings.Builder
+		for i := range values {
```
yogilad, line 199: Q: How come values is a range on top of headers?

yogilad (approved):
Question in comments

kusto/conn.go
AsafMah: They made it so headers can be repeated, so each header is an array of strings (if that's what you meant)

yogilad: Thanks!

PR #185, opened 2023-05-01
fix: Assign ClientOptions if provided

JorTurFer:
Currently, the function `AttachPolicyClientOptions` doesn't do anything because the condition is wrong. Only nil options are assigned, which is not correct.
This PR solves the problem


JorTurFer:
FYI @AsafMah

AsafMah:
Wow, great fix.
I'll release a patch soon.

AsafMah (approved):

JorTurFer:
> I'll release a patch soon.

Do you have an ETA? We need this fix in KEDa

AsafMah:
Probably tommorow, You can refer directly to the master branch in the meantime though

AsafMah:
Fixed in 0.13.0

PR #187, opened 2023-05-02
Remove MgmtOption

AsafMah:
### Added

- `ServerTimeout` Query Option

### Changed

- `MgmtOption` is deprecated. From now on both `Query` and `Mgmt` accept `QueryOption`, `MgmtOption` will remain as an
  alias until the next version.

### Removed

- `AllowWrite` has been a no-op for a while. It is now finally removed.

### Security


CHANGELOG.md
```diff
@@ -1,110 +1,178 @@
...
+
+### Added
+
+- `ServerTimeout` Query Option
+
+### Changed
+
+- `MgmtOption` is deprecated. From now on both `Query` and `Mgmt` accept `QueryOption`, `MgmtOption` will remain as an
+  alias until the next version.
+
```
yogilad, line 18: Setting a time based ctx does not limit server time.
Use ServerTimeout instead

kusto/queryopts.go
```diff
@@ -135,14 +135,9 @@ func ResultsProgressiveDisable() QueryOption {
...
 }
 
-// queryServerTimeout is the amount of time the server will allow a query to take.
-// NOTE: I have made the serverTimeout private. For the moment, I'm going to use the context.Context timer
-// to set timeouts via this private method.
-func queryServerTimeout(d time.Duration) QueryOption {
+// ServerTimeout overrides the default request timeout.
+func ServerTimeout(d time.Duration) QueryOption {
 	return func(q *queryOptions) error {
-		if d > 1*time.Hour {
```
yogilad, line 143: Why?
Is the service contract different?

AsafMah: We just don't have this limit in any other SDK, so I removed it for parity

ohadbitt: In other SDKs though we take the timeout from the additional properties

yogilad (approved):

ohadbitt (approved):

PR #190, opened 2023-05-24
Don't add timeout to context

AsafMah:


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+  * Fixed wrong context deadline setting
```
yogilad, line 11: This being a somewhat breaking change, it's worth a better description about the relation between context and request_timeout

AsafMah: The behavior is explained in the 13.1 release.
This was a bug causing the behavior to be different, so by fixing it does align with what's written before.

yogilad: You mean this _"Modified once.go to reset sync.Once instance when an error occurs"_
1. This is very technical 
2. How is this related?

AsafMah: I meant  
```
## [0.13.0] - 2023-05-09
### Added
- `ServerTimeout` Query Option
  - The timeout to the server will be set to the value of this option, or to none if `RequestNoTimeout` is set to true.
  - If it is not provided, the timeout will be set by the context (the old behaviour).
  - If a context timeout is not provided, it will fall back to a default value by the type of request.
```

PR #191, opened 2023-05-28
Support urls with dns zone

ohadbitt:
Also renamed trusted_endpoints package to trustedendpoints as this change sat in my env for a while after not pushing it


kusto/test/etoe/etoe_test.go
```diff
@@ -20,6 +15,12 @@ import (
 	"time"
 	"unicode"
 
+	"github.com/Azure/azure-kusto-go/kusto/utils"
```
AsafMah, line 18: why?

kusto/ingest/internal/resources/resources.go
```diff
@@ -63,12 +64,23 @@ func parse(uri string) (*URI, error) {
...
 	}
 
-	v := &URI{
-		u:          u,
-		account:    hostSplit[0],
-		objectType: hostSplit[1],
-		objectName: strings.TrimLeft(u.EscapedPath(), "/"),
-		sas:        u.Query(),
+	var v *URI
+	if len(hostSplit) == 5 {
```
AsafMah, line 68: that's the only way to check? the zone can't happen to have 5 segments? is it what we do in other sdks?

AsafMah (changes requested):

kusto/test/etoe/etoe_test.go
ohadbitt: revert

kusto/ingest/internal/resources/resources.go
ohadbitt: This is what we do for example- you see we use the objectName

	
	return azqueue.NewServiceURL(*service, p).NewQueueURL(queue.ObjectName()).NewMessagesURL(), nil

AsafMah: https://github.com/Azure/azure-storage-blob-go is deprecated - see the readme (if that's what you meant)

I think they might have added methods to make it a bit easier recently.

Either way I'll resolve for now

PR #192, opened 2023-06-10
User/ohbitton/streaming ingest from blob

ohadbitt:
# Added
Streaming ingest from blob


kusto/ingest/file_options.go
```diff
@@ -459,3 +459,17 @@ func ClientRequestId(clientRequestId string) FileOption {
 		name:         "ClientRequestId",
 	}
 }
+
+// RawDataSize is the uncompressed data size. Should be used to comunicate the file size to the service for efficient ingestion.
+// Also used by managed client in the decision to use queued ingestion instead of streaming (if > 4mb)
```
yogilad, line 464: But, following our discussion, this is not true.
The compressed estimate is the one with the 4MB limit, not the raw one.

kusto/ingest/internal/resources/resources.go
```diff
@@ -395,3 +399,26 @@ func InitBackoff() backoff.BackOff {
 	exp.Multiplier = defaultMultiplier
 	return backoff.WithMaxRetries(exp, retryCount)
 }
+
+func FetchBlobSize(fPath string, ctx context.Context, client *http.Client) (size int64, err error) {
```
yogilad, line 403: Why is this an API of resource manager?
Also mind,
* We support semicolon credentials which AzBlob does not.
* We support MI auth which can not be run locally.
* We support storages like S3 we do not calculate size to

In all these cases, need to decide if we want to proceed to streaming or fallback to queued immediately when a blob size is not provided.

kusto/ingest/internal/resources/resources.go
```diff
@@ -45,8 +49,8 @@ type URI struct {
 	sas                             url.Values
 }
 
-// parse parses a string representing a Kutso resource URI.
-func parse(uri string) (*URI, error) {
+// Parse parses a string representing a Kutso resource URI.
+func Parse(uri string) (*URI, error) {
```
yogilad, line 53: I think the API can be tested without being made public

kusto/ingest/managed.go
```diff
@@ -44,11 +46,72 @@ func NewManaged(client QueryClient, db, table string, options ...Option) (*Manag
...
 	props := m.newProp()
 	file, err := prepFileAndProps(fPath, &props, options, ManagedClient)
 
-	if err == FileIsBlobErr { // Non-local file - fallback to queued
+	if file == nil {
+		if props.Ingestion.RawDataSize == 0 {
+			size, err := resources.FetchBlobSize(fPath, ctx, m.queued.client.HttpClient())
+			if err != nil {
+				// Failed fetch blob properties
+				return nil, err
```
yogilad, line 103: You should consider queue or try none the less first.
Definitely not fail the operation

kusto/ingest/managed.go
```diff
@@ -44,11 +46,72 @@ func NewManaged(client QueryClient, db, table string, options ...Option) (*Manag
...
+			}
+			props.Ingestion.RawDataSize = size
+		}
+
+		if props.Ingestion.RawDataSize <= maxStreamingSize {
```
yogilad, line 108: You are not taking into account compression.

kusto/ingest/managed.go
```diff
@@ -44,11 +46,72 @@ func NewManaged(client QueryClient, db, table string, options ...Option) (*Manag
...
+
+	return nil, err
+}
+
 func (m *Managed) FromFile(ctx context.Context, fPath string, options ...FileOption) (*Result, error) {
 	props := m.newProp()
 	file, err := prepFileAndProps(fPath, &props, options, ManagedClient)
 
-	if err == FileIsBlobErr { // Non-local file - fallback to queued
+	if file == nil {
```
yogilad, line 98: Do you mean _if err = nil {_ ?

kusto/ingest/managed.go
```diff
@@ -44,11 +46,72 @@ func NewManaged(client QueryClient, db, table string, options ...Option) (*Manag
...
+	hasCustomId := props.Streaming.ClientRequestId != ""
+	i := 0
+	managedUuid := uuid.New().String()
+
+	actualBackoff := backoff.WithContext(backoff.WithMaxRetries(props.ManagedStreaming.Backoff, retryCount), ctx)
+
+	var err error = nil
+	err = backoff.Retry(func() error {
+		if !hasCustomId {
+			props.Streaming.ClientRequestId = fmt.Sprintf("KGC.executeManagedStreamingIngest;%s;%d", managedUuid, i)
```
yogilad, line 64: This can be done once

kusto/ingest/internal/resources/resources.go
```diff
@@ -395,3 +399,26 @@
 	exp.Multiplier = defaultMultiplier
 	return backoff.WithMaxRetries(exp, retryCount)
 }
+
+func FetchBlobSize(fPath string, ctx context.Context, client *http.Client) (size int64, err error) {
```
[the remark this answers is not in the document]
ohadbitt: This refers to blob connection string (you know -something from this build
DefaultEndpointsProtocol=http;AccountName=acc;
AccountKey=key

kusto/conn_streaming_ingest.go
```diff
@@ -51,7 +55,9 @@
 	properties.ClientRequestID = clientRequestId
 	headers := c.getHeaders(properties)
 	headers.Del("Content-Type")
-	headers.Add("Content-Encoding", "gzip")
+	if !isBlobUri {
+		headers.Add("Content-Encoding", "gzip")
```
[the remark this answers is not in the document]
ohadbitt: I think not - this is only streaming ingestion - we either get it gzipped or gzip it

kusto/ingest/file_options.go
ohadbitt: That was before this decision and is currently true - we dont have any estimation in the go client so this will change will have to be in a new version together with all the other sdks

kusto/ingest/internal/resources/resources.go
ohadbitt: interesting - this is again relevant to all sdks and need to be discussed -

kusto/ingest/internal/resources/resources.go
ohadbitt: its not for testing - i use it in streaming package now as well

kusto/ingest/managed.go
ohadbitt: as said above - this needs to be done in another pr

kusto/ingest/managed.go
ohadbitt: No

kusto/ingest/managed.go
ohadbitt: I disagree
This is something we do all the time also in queued ingestion - if fetch blob size fails we throw - why should we think that the DM would be able to find the blob if we failed ?

kusto/ingest/internal/resources/resources.go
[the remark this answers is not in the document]
AsafMah: You are right.
But I found a new method -azblob.ParseURL (which turns out we already use elsewhere in the code), which at least does the parsing for us (we still have to create the client).

kusto/ingest/managed.go
ohadbitt: we add the retry number to it

kusto/ingest/internal/resources/resources.go
yogilad: discussed f2f

kusto/ingest/managed.go
yogilad: discussed f2f

kusto/ingest/managed.go
yogilad: discussed f2f

kusto/conn_streaming_ingest.go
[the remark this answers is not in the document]
ohadbitt: Decided with @yogilad that we will document this as problematic 
I changed the docstring to :
In streaming - do not pass DontCompress if file is not already compressed.

kusto/ingest/file_options.go
ohadbitt: After discussion - Done in this version

kusto/ingest/managed.go
ohadbitt: After discussion - decided to fetch size for keys as well and return 0 for ManagedIdentity or amazon

kusto/ingest/managed.go
ohadbitt: changed

kusto/ingest/streaming.go
```diff
@@ -50,48 +58,58 @@ func NewStreaming(client QueryClient, db, table string) (*Streaming, error) {
...
-func prepFileAndProps(fPath string, props *properties.All, options []FileOption, client ClientScope) (*os.File, error) {
-	local, err := queued.IsLocalPath(fPath)
 	if err != nil {
 		return nil, err
 	}
 
+	return streamImpl(i.streamConn, ctx, file, props, false)
+}
+
+func prepFileAndProps(fPath string, props *properties.All, options []FileOption, client ClientScope) (*os.File, error, bool) {
```
AsafMah, line 78: Mmmm, not sure how I feel about the third argument.
Maybe we'll have os.File just be nil if it's not local?

kusto/ingest/managed.go
```diff
@@ -44,11 +46,72 @@ func NewManaged(client QueryClient, db, table string, options ...Option) (*Manag
 	}, nil
 }
 
+// Attempts to stream with retries, on success - return res,nil.
+// If failed permanently - return err,nil.
+// If failed transiently - return nil,nil.
+func (m *Managed) streamWithRetries(ctx context.Context, payloadProvider func() io.Reader, props properties.All, isBlobUri bool) (*Result, error) {
```
AsafMah, line 52: Why is payloadprovider a function rather than just an argument?

AsafMah: Also do you close it? need to close it?

kusto/ingest/internal/utils/ingestion_utils.go
```diff
@@ -0,0 +1,110 @@
...
+
+	parsed, err := resources.Parse(fPath)
+	if err != nil {
+		return 0, err
+	}
+
+	var blobClient *azblob.Client = nil
+	var objectNameSplit []string
+
+	if len(parsed.SAS()) == 0 {
```
AsafMah, line 37: Do we really not have shared code for any of this?

kusto/ingest/streaming.go
ohadbitt: This was what i had initially and Yocahi said its not readable

kusto/ingest/managed.go
ohadbitt: See line 166 
	res, err := m.streamWithRetries(ctx, func() io.Reader { return bytes.NewReader(buf) }, props, false)

Reader interface has only Read() no close ...

kusto/ingest/streaming.go
yogilad: A doc string would help clarify the meaning of each returned value

kusto/ingest/managed.go
```diff
@@ -93,25 +93,25 @@ func (m *Managed) streamWithRetries(ctx context.Context, payloadProvider func()
...
-		if props.Ingestion.RawDataSize == 0 {
-			size, err := resources.FetchBlobSize(fPath, ctx, m.queued.client.HttpClient())
-			if err != nil {
-				// Failed fetch blob properties
-				return nil, err
-			}
-			props.Ingestion.RawDataSize = size
+	file, err, local := prepFileAndProps(fPath, &props, options, ManagedClient)
+
+	if !local {
```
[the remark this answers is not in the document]
ohadbitt: added commetn

PR #195, opened 2023-06-15
Remove from URLs query parameters that might have secrets

yihezkel:
### Security
Remove query parameters from URL when they might have secrets and might be traced (even if part of error flow)


AsafMah (approved):
You can remove the failed test, testing against the exact JSON output is bad anyway, and not related to your change.

kusto/ingest/file_options.go
```diff
@@ -219,9 +219,9 @@ const (
 	SingleJSON DataFormat = properties.SingleJSON
 )
 
-// InferFormatFromFileName  looks at the file name and tries to discern what the file format is
+// InferFormatFromFileName looks at the file name and tries to discern what the file format is
 func InferFormatFromFileName(fName string) DataFormat {
-	return properties.DataFormatDiscovery(fName)
+	return properties.DataFormatDiscovery(properties.RemoveQueryParamsFromUrl(fName))
```
yogilad, line 224: How come this is relevant here?

kusto/ingest/ingest.go
```diff
@@ -139,7 +139,7 @@ func (i *Ingestion) FromFile(ctx context.Context, fPath string, options ...FileO
 
 // fromFile is an internal function to allow managed streaming to pass a properties object to the ingestion.
 func (i *Ingestion) fromFile(ctx context.Context, fPath string, options []FileOption, props properties.All) (*Result, error) {
-	local, err := queued.IsLocalPath(fPath)
+	local, err := queued.IsLocalPath(properties.RemoveQueryParamsFromUrl(fPath))
```
yogilad, line 142: How come this is relevant here?

kusto/ingest/internal/queued/queued.go
```diff
@@ -219,7 +219,7 @@ func (i *Ingestion) Blob(ctx context.Context, from string, fileSize int64, props
 
 	props.Ingestion.RetainBlobOnSuccess = !props.Source.DeleteLocalSource
 
-	err = CompleteFormatFromFileName(&props, from)
+	err = CompleteFormatFromFileName(&props, properties.RemoveQueryParamsFromUrl(from))
```
yogilad, line 222: Can we do this only where we return an error or an object where the full URI is not needed?

kusto/ingest/internal/properties/properties.go
```diff
@@ -416,6 +416,13 @@ func (i Ingestion) validate() error {
 	return nil
 }
 
+func RemoveQueryParamsFromUrl(url string) string {
```
yogilad, line 419: it might be easier to use if the API was `func (s string) RemoveQueryParamsFromUrl() string`
calling would simply be s.RemoveQueryParamsFromUrl()

kusto/ingest/streaming.go
```diff
@@ -59,7 +59,7 @@ func (i *Streaming) FromFile(ctx context.Context, fPath string, options ...FileO
 }
 
 func prepFileAndProps(fPath string, props *properties.All, options []FileOption, client ClientScope) (*os.File, error) {
-	local, err := queued.IsLocalPath(fPath)
+	local, err := queued.IsLocalPath(properties.RemoveQueryParamsFromUrl(fPath))
```
yogilad, line 62: same

kusto/ingest/streaming.go
```diff
@@ -82,7 +82,7 @@ func prepFileAndProps(fPath string, props *properties.All, options []FileOption,
 		props.Source.DontCompress = true
 	}
 
-	err = queued.CompleteFormatFromFileName(props, fPath)
+	err = queued.CompleteFormatFromFileName(props, properties.RemoveQueryParamsFromUrl(fPath))
```
yogilad, line 85: same

kusto/ingest/file_options.go
yihezkel: Every place I did this, it's a place where:
1) Secrets are unnecessary for the called function to work.
2) The function (perhaps transitively) can throw an error when doing something with the URL. I'm less familiar with Go's error handling, but my concern is that the exception may include the parameter.

kusto/ingest/internal/properties/properties.go
yihezkel: I'm definitely not an expert in Go, but I tried that because it was my preferred approach as well, and it didn't work. IIRC the issue was that I couldn't add a function on the `string` class.

IIRC we do have a 'String` class, and I tried to add this method to there, but I believe it would've required more extensive changes to pass that type of parameter instead of the existing `string`.

AsafMah: Yihezkel is right, you can't monkeypatch classes on go (you can't do anything fun on go).
The capital String class is for the string type in kusto (parsing them, escaping them, etc), and should not be used for general purpose.

kusto/ingest/internal/queued/queued.go
yogilad: This is not the right way to enforce this.
If you want to do something covering all possible paths, create a SecretiveUri like structure and have String() and SecureString() differ on the output.

AsafMah: Not 100% sure

I get the idea, and type safety is good, but I wonder if it can even be done correctly without infecting the whole code and having breaking changes

kusto/ingest/internal/properties/properties.go
```diff
@@ -416,6 +416,13 @@ func (i Ingestion) validate() error {
...
+	if idx := strings.Index(url, "?"); idx != -1 {
```
ohadbitt, line 420: add also 
strings.Index(url, ";")

kusto/ingest/internal/queued/queued.go
yihezkel: As discussed, I'm going to reduce the scope of this to only the explicit prints.

Afterwards, when I have time, I'll pass an object that contains both a secretive and a sanitized URL, and pass the latter to the methods that don't require the secret.

AsafMah (approved):

PR #196, opened 2023-06-19
Add workload identity

PheonixS:
### Added
Kubernetes Workload Identity from azidentity library.
### Changed
Bumped version of azidentity from v1.2.0 to v1.3.0


PheonixS:
@microsoft-github-policy-service agree

OlehPriadka-TomTom (approved):

OlehPriadka-TomTom (changes requested):
approve by mistake

AsafMah:
WithKubernetesWorkloadIdentity() should accept the parameters it needs, instead of assuming they are in the string.
If the parameters are optional - just mention it in the docs.

kusto/kcsb.go
```diff
@@ -251,6 +254,15 @@ func (kcsb *ConnectionStringBuilder) WithSystemManagedIdentity() *ConnectionStri
 	return kcsb
 }
 
+// WithKubernetesWorkloadIdentity Creates a Kusto Connection string builder that will authenticate with AAD application, using
+// an application token obtained from a Microsoft Service Identity endpoint using Kubernetes workload identity.
+func (kcsb *ConnectionStringBuilder) WithKubernetesWorkloadIdentity() *ConnectionStringBuilder {
```
yogilad, line 259: The API ought to take in any argument that is required to configure this auth mode. 
This one seems to require 3 (app id, token file, authority id)

kusto/kcsb.go
```diff
@@ -23,6 +24,8 @@ type ConnectionStringBuilder struct {
 	ApplicationToken                 string
 	AzCli                            bool
 	MsiAuthentication                bool
+	WorkloadAuthentication           bool
+	FederationTokenFile              string
```
yogilad, line 28: Is this the file content or the file path?
if the latter, please add "Path" to the member name.

kusto/kcsb_test.go
```diff
@@ -166,6 +166,12 @@ func TestGetTokenProviderHappy(t *testing.T) {
 				DataSource:        "https://endpoint/test_tokenprovider_managedidauth2",
 				MsiAuthentication: true,
 			},
+		}, {
```
yogilad, line 169: Please also add an API test for WithKubernetesWorkloadIdentity(...)

kusto/kcsb.go
yogilad: This ought to also affect tests and code example.

yogilad (commented):
Thanks for the contribution.
I left comments in the PR.

kusto/kcsb.go
PheonixS: done

kusto/kcsb_test.go
PheonixS: done

kusto/kcsb.go
PheonixS: done

AsafMah (approved):

PR #201, opened 2023-08-17
kusto: fix dropped error

alrs:
This fixes a dropped `err` variable in the `kusto` package.

I labeled the error of the sort `errors.OpQuery`. If I should have chosen something else I'm happy to make the change.


AsafMah:
Great Catch, thanks!

PR #205, opened 2023-09-19
Use klauspost/gzip instead of compress/gzip

jwilder:
### Changed

Switches uploading with gzip to use https://github.com/klauspost/compress.  This can result in 30%-50% reduction in compression time (depending on the shape of the data).  In our setup, we're upload many files continuously and this reduces the CPU utilization on the hosts allowing for greater throughput with less latency.


AsafMah:
Thanks for the effort, but we rather not include extra deps unless it's critical.
You can always pass files or streamed that are already compressed by whichever way you want.
If it can't be inferred from the extension, you can explicitly pass the DontCompress option to source.

PR #214, opened 2023-11-08
Report size of original reader - not the compressed size

ohadbitt:
### Fixed
-     Size used for RawDataSize taken from gzip reader was of the gzip size and not the original reader size

### DIFF
--- a/kusto/ingest/internal/gzip/gzip.go
+++ b/kusto/ingest/internal/gzip/gzip.go
@@ -43,7 +43,7 @@ func (s *Streamer) Reset(reader io.ReadCloser) {
 // InputSize returns the amount of data that the Streamer streamed. This will only be accurate for
 // the full stream after Read() has returned io.EOF and not before.
 func (s *Streamer) InputSize() int64 {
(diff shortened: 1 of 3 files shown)


kusto/ingest/internal/gzip/gzip.go
```diff
@@ -69,7 +69,9 @@ func (s *Streamer) run() {
 		defer zw.Close()
 		defer zw.Flush()
 
-		_, err := io.Copy(zw, s.userInput)
+		amount, err := io.Copy(zw, s.userInput)
+		atomic.AddInt64(&s.size, int64(amount))
```
yogilad, line 73: Why Add?
Can this be run on multiple streams?

yogilad (approved):
with comment

kusto/ingest/internal/gzip/gzip.go
ohadbitt: Yea i guess you are right - it was before add because we did the chunking but here the copy does it for us

AsafMah (approved):

PR #216, opened 2023-11-26
Query v2

AsafMah:


azkustodata/kql/query_parameters.go
```diff
@@ -1,70 +1,74 @@
 package kql
 
 import (
-	"github.com/Azure/azure-kusto-go/azkustodata/types"
```
yogilad, line 4: Why the rename?
The old name sounds more to point.

azkustodata/kql/query_parameters.go
```diff
@@ -1,70 +1,74 @@
...
 		panic("Invalid parameter values. make sure to adhere to KQL entity name conventions and escaping rules.")
 	}
-	q.parameters[key] = value
+	q.parameters[key] = v
 	return q
 }
 
-func (q *Parameters) AddBool(key string, value bool) *Parameters {
-	return q.addBase(key, newValue(value, types.Bool))
+func (q *Parameters) AddBool(key string, v bool) *Parameters {
```
yogilad, line 30: Same question - does this support nulls?

azkustodata/kusto.go
```diff
@@ -12,12 +15,14 @@ import (
...
 )
 
+type Statement = *kql.Builder
+
 // queryer provides for getting a stream of Kusto frames. Exists to allow fake Kusto streams in tests.
 type queryer interface {
 	io.Closer
 	query(ctx context.Context, db string, query Statement, options *queryOptions) (execResp, error)
 	mgmt(ctx context.Context, db string, query Statement, options *queryOptions) (execResp, error)
-	queryToJson(ctx context.Context, db string, query Statement, options *queryOptions) (string, error)
```
yogilad, line 20: Why break this APIs Name?

azkustodata/kusto.go
```diff
@@ -172,6 +177,86 @@ func (c *Client) Query(ctx context.Context, db string, query Statement, options
 	return iter, nil
 }
 
+func (c *Client) newV1(
```
yogilad, line 180: New is usually reserved for instance creation.
Why do you use it in query/command APIs?

The set of API names need working.
Possible [Mgmt, QueryV1, QueryV2, IterativeQuery] where Iterative is V2 implicitly.

azkustodata/kusto.go
```diff
@@ -172,6 +177,86 @@ func (c *Client) Query(ctx context.Context, db string, query Statement, options
...
+
+	return queryv2.NewStreamingDataSet(ctx, res, capacity)
+}
+
+func (c *Client) newV2(ctx context.Context, db string, kqlQuery Statement, options []QueryOption) (*queryOptions, io.ReadCloser, error) {
+	ctx, cancel := contextSetup(ctx)
+
+	options = append(options, V2NewlinesBetweenFrames())
+	options = append(options, V2FragmentPrimaryTables())
+	options = append(options, ResultsErrorReportingPlacement(ResultsErrorReportingPlacementEndOfTable))
```
yogilad, line 238: Do you test that being last allows you to override these options no matter what the user did?
Any options conflicting with those you should forcibly disable or fail the query over?

azkustodata/kusto.go
```diff
@@ -172,6 +177,86 @@ func (c *Client) Query(ctx context.Context, db string, query Statement, options
...
+		return nil, nil, err
+	}
+
+	res, err := conn.rawQuery(ctx, db, kqlQuery, opts)
+
+	if err != nil {
+		cancel()
+		return nil, nil, err
+	}
+	return opts, res, nil
```
yogilad, line 257: Where's the whole iterative dataset reader?

azkustodata/kusto.go
```diff
@@ -172,6 +177,86 @@ func (c *Client) Query(ctx context.Context, db string, query Statement, options
...
+}
+
 func (c *Client) QueryToJson(ctx context.Context, db string, query Statement, options ...QueryOption) (string, error) {
```
yogilad, line 260: Should you not leave this API public?
People may have taken a dependency on it?

azkustodata/kusto.go
```diff
@@ -200,12 +290,6 @@ func (c *Client) QueryToJson(ctx context.Context, db string, query Statement, op
 // Note that the server has a timeout of 10 minutes for a management call by default unless the context deadline is set.
 // There is a maximum of 1 hour.
 func (c *Client) Mgmt(ctx context.Context, db string, query Statement, options ...QueryOption) (*RowIterator, error) {
-	if stmt, ok := query.(Stmt); ok {
```
yogilad, line 203: Did this move somewhere?
Otherwise, why the change?

azkustodata/query/column.go
```diff
@@ -0,0 +1,13 @@
+package query
+
+import "github.com/Azure/azure-kusto-go/azkustodata/types"
+
+// Column represents a column in a table.
+type Column interface {
```
yogilad, line 6: Why would such a simple class need an interface?

azkustodata/query/from_kusto.go
```diff
@@ -0,0 +1,66 @@
+package query
+
+// value.go provides methods for converting a row to a *struct and for converting KustoValue into Go types
```
yogilad, line 3: update this (I did not look at other files, but they can also be out of date)

yogilad (commented):
very partial review (read through some files)
Will carry on later

azkustodata/example_iterative_query_test.go
```diff
@@ -0,0 +1,175 @@
...
+		if err != nil {
+			panic(err)
+		}
+	}(client)
+
+	ctx := context.Background()
+
+	// Simple query - single table
+
+	dataset, err := client.IterativeQuery(ctx, "Samples", kql.New("PopulationData"))
```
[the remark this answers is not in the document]
AsafMah: This table was chosen because it only has 52 rows

azkustodata/example_mgmt_test.go
```diff
@@ -0,0 +1,112 @@
...
+		if err != nil {
+			panic(err)
+		}
+	}(client)
+
+	ctx := context.Background()
+
+	// Simple query - single table
+
+	dataset, err := client.MgmtNew(ctx, "Samples", kql.New(".show tables"))
```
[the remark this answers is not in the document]
AsafMah: Mgmt is taken, it's there until I remove the old one.

azkustodata/example_mgmt_test.go
```diff
@@ -0,0 +1,112 @@
...
+		println(row.Ordinal())
+
+		// Or iterate over all rows
+		for _, row := range rows {
+			// Each row has an index and a pointer to the table it belongs to
+			println(row.Ordinal())
+			println(row.Table().Name())
+
+			// There are a few ways to access the values of a row:
+			val := row.Value(0)
```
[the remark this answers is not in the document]
AsafMah: What's GetTypeByRef? The Value object has a "GetType" method if this is what you mean?

azkustodata/kql/builder.go
```diff
@@ -43,6 +43,11 @@ func (b *Builder) addBase(value fmt.Stringer) *Builder {
 	return b
 }
 
+func (b *Builder) AddValue(v value.Kusto) *Builder {
```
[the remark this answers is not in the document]
AsafMah: The name was here before me.  It means "Kusto Value" I guess.
We can call it value.Value or another name if you have it

azkustodata/kql/builder.go
```diff
@@ -55,44 +60,48 @@ func (b *Builder) AddLiteral(value stringConstant) *Builder {
 	return b.addBase(value)
 }
 
-func (b *Builder) AddBool(value bool) *Builder {
-	return b.addBase(newValue(value, types.Bool))
+func (b *Builder) AddBool(v bool) *Builder {
```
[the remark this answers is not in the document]
AsafMah: It doesn't.
The type bool doesn't support it, I can change it to accept a bool reference, but that isn't great either.

There is now an "AddValue" api you can add from a (nullable) kusto value:
b.AddValue(value.NewNullBool())

azkustodata/kql/query_parameters.go
AsafMah: If you have a "types.Bool" you don't have a bool type, you have a bool value. 
That was my logic, but I can revert it otherwise.

azkustodata/kql/query_parameters.go
AsafMah: Same answer

azkustodata/kusto.go
AsafMah: 1. This is a private api
2. The method no longer returns a json string - it returns a raw response object.

azkustodata/kusto.go
AsafMah: all of the "major" code is in query.v2

azkustodata/kusto.go
AsafMah: This was an api that was specifically requested from us

azkustodata/query/column.go
AsafMah: To not have breaking changes if the internals change.

azkustodata/example_mgmt_test.go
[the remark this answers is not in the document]
yogilad: Sorry, Get[Type]By[Ref]() as in the set of APIs for _GetBooleanByOrdinal_, _GetStringByColumnName_ and so on

azkustodata/kql/builder.go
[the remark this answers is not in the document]
yogilad: What does this do?
Isn't this the interface for all KQL types?

azkustodata/example_mgmt_test.go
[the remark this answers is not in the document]
yogilad: New implies a CTOR like behavior in Go.
Can you add a comment this is temporary and should not be committed?

azkustodata/kql/builder.go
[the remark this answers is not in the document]
yogilad: Why is that not great?
What happens when you get a type back? how do you indicate a row is null?

azkustodata/kql/query_parameters.go
yogilad: not sure I follow.
Those are data types. the do store values, but you code the type behavior.
Suggest to revert (it would also simplify this PR)

azkustodata/kusto.go
yogilad: Not sure how this answers the question.
Can you clarify?

yogilad (commented):
Just replies

azkustodata/kusto.go
AsafMah: Not sure what you mean then - I didn't change this API

azkustodata/kusto.go
AsafMah: It will override all of the options.
I can make it fail if they define progressive=true if you want

azkustodata/kusto.go
AsafMah: Again, the conflict with the old names

azkustodata/kql/builder.go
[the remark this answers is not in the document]
AsafMah: Because they now can't do AddBool(true), they have to define a variable and make it into a reference.

When returning it's simpler

azkustodata/kql/query_parameters.go
AsafMah: My mistake - I didn't change any namespace, they were named like this before. 
the types package defines the column types (like types.Bool is a constant that means "bool"), and the value defines a value (like value.Bool is a struct that holds a kusto bool ).

The difference here is now I need a different improt

azkustodata/kusto.go
AsafMah: not sure why, I'll revert - but this whole method is going to be removed soon anyway.

azkustodata/query/from_kusto.go
AsafMah: It's only for this file, which was taken from the old system.

azkustodata/example_iterative_query_test.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	// The tables are streamed, so you need to iterate through the `Results()` channel to get them
+
+	for tableResult := range dataset.Results() {
+		// Make sure to always check for errors
+		if tableResult.Err() != nil {
+			// It's possible to get an errors and still get a table - partial results
```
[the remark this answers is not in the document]
AsafMah: In fatal errors there will be no table results.

azkustodata/example_iterative_query_test.go
```diff
@@ -0,0 +1,175 @@
...
+		// 2. GetAllRows() - reads all rows and closes the table
+		rows, errors := table.GetAllRows()
+		for _, row := range rows {
+			println(row.Ordinal())
+		}
+		for _, err := range errors {
+			println(err.Error())
+		}
+
+		// 3. Rows() - reads rows as they are received
```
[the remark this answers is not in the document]
AsafMah: Fixed

azkustodata/example_iterative_query_test.go
```diff
@@ -0,0 +1,175 @@
...
+			err = row.ToStruct(&pd)
+			if err != nil {
+				panic(err)
+			}
+			println(pd.State)
+			println(pd.Pop)
+
+		}
+
+		// Or you can easily get the results as a slice of structs
```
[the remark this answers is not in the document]
AsafMah: Fixed

azkustodata/example_mgmt_test.go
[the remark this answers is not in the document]
AsafMah: Added

azkustodata/example_mgmt_test.go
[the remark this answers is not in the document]
AsafMah: Oh yes, I'll add the example

azkustodata/kql/builder.go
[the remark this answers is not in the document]
AsafMah: Yes, what would you call it?

azkustodata/query/v2/iterative_dataset.go
```diff
@@ -0,0 +1,123 @@
...
+		rows, err := table.GetAllRows()
+		tables = append(tables, query.NewDataTable(d, table.Ordinal(), table.Id(), table.Name(), table.Kind(), table.Columns(), rows, err))
+	}
+
+	return tables, errs
+}
+
+// NewStreamingDataSet creates a new iterativeDataset from a reader.
+// The capacity parameter is the capacity of the channel that receives the frames from the Kusto service.
+func NewStreamingDataSet(ctx context.Context, r io.ReadCloser, capacity int) (IterativeDataset, error) {
```
yogilad, line 104: Itraerative

azkustodata/query/v2/json.go
```diff
@@ -0,0 +1,209 @@
...
+
+// readFramesIterative reads frames from a reader and sends them to a channel as they are read.
+func readFramesIterative(reader io.Reader, ch chan<- Frame) error {
+	defer close(ch)
+
+	// Crazily enough, json.Decoder always puts THE ENTIRE READER IN MEMORY
+	// So we have to manually split the reader into lines and decode each line with a new decoder
+
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
```
yogilad, line 140: See if this duplicates the data from the input stream?
What does slice do?

azkustodata/value/value.go
```diff
@@ -35,15 +35,47 @@ The Unmarshal() is for internal use, it should not be needed by an end user. Use
...
+	case types.Int:
+		return NewNullInt()
+	case types.Long:
+		return NewNullLong()
+	case types.Real:
+		return NewNullReal()
+	case types.Decimal:
+		return NewNullDecimal()
+	case types.String:
+		return NewNullString()
```
yogilad, line 67: No null in string

azkustodata/query/table.go
```diff
@@ -0,0 +1,29 @@
+package query
+
+import "github.com/Azure/azure-kusto-go/azkustodata/errors"
+
+type Table interface {
+	Id() string
+	Index() int64
+	Name() string
+	Columns() []Column
+	Kind() string
```
yogilad, line 10: enum or consts

azkustodata/query/table_impl.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func NewFullTable(base Table, rows []Row) FullTable {
+	return &fullTable{
+		Table: base,
+		rows:  rows,
+	}
+}
+
+func ReplaceFullTableRows(t FullTable, rows []Row) FullTable {
```
yogilad, line 85: reminder: row ref to col instead of table and maybe this will become redundant

azkustodata/query/v2/full_dataset.go
```diff
@@ -0,0 +1,40 @@
+package v2
+
+import (
+	"context"
+	"github.com/Azure/azure-kusto-go/azkustodata/errors"
+	"github.com/Azure/azure-kusto-go/azkustodata/query"
+)
+
+// FullDataset represents a full result from kusto - where all the tables are received before the dataset is returned.
+type fullDataset struct {
```
yogilad, line 10: Check if this is a duplicate of the shared impl

PR #217, opened 2023-12-04
Improve stability by selecting storage resources based on success rates.

alonadam:
### Added
Rank storage resources by usage statistics.

### DIFF
--- a/kusto/ingest/internal/queued/queued.go
+++ b/kusto/ingest/internal/queued/queued.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"math/rand"
 	"net/http"
 	"net/url"
 	"os"
@@ -35,8 +34,9 @@ const (
 	// made a 5x improvement in speed. We don't have any numbers from the service side to give us numbers we should use, so this
 	// is our best guess from observation. DO NOT CHANGE UNLESS YOU KNOW BETTER.
 
-	BlockSize   = 8 * _1MiB
-	Concurrency = 50
+	BlockSize             = 8 * _1MiB
+	Concurrency           = 50
+	StorageMaxRetryPolicy = 3
 )
 
 // Queued provides methods for taking data from various sources and ingesting it into Kusto using queued ingestion.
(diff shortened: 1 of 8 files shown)


kusto/ingest/internal/resources/ranked_storage_account.go
```diff
@@ -0,0 +1,95 @@
+package resources
+
+type StorageAccountStats struct {
+	successCount int
+	totalCount   int
+}
+
+func NewStorageAccountStats() *StorageAccountStats {
```
AsafMah, line 8: This is never used, you call the manual init

kusto/ingest/internal/resources/ranked_storage_account.go
```diff
@@ -0,0 +1,95 @@
...
+	currentBucketIndex int
+	accountName        string
+	numberOfBuckets    int
+	bucketDuration     int64
+	timeProvider       func() int64
+}
+
+func newRankedStorageAccount(accountName string, numberOfBuckets int, bucketDuration int64, timeProvider func() int64) *RankedStorageAccount {
+	buckets := make([]StorageAccountStats, numberOfBuckets)
+	for i := range buckets {
```
AsafMah, line 39: You actually don't need this loop

kusto/ingest/internal/resources/resources.go
```diff
@@ -44,7 +44,7 @@ type URI struct {
 	sas                             url.Values
 }
 
-// Parse parses a string representing a Kutso resource URI.
+// Pparses a string representing a Kutso resource URI.
```
AsafMah, line 47: oops?

kusto/ingest/internal/resources/ranked_storage_account_set.go
```diff
@@ -0,0 +1,104 @@
...
+	defaultNumberOfBuckets         = 6
+	defaultBucketDurationInSeconds = 10
+)
+
+var defaultTiersValue = [4]int{90, 70, 30, 0}
+var defaultTimeProvider = func() int64 { return time.Now().Unix() }
+
+type RankedStorageAccountSet struct {
+	accounts          map[string]*RankedStorageAccount
+	number_of_buckets int
```
AsafMah, line 19: camelCase

kusto/ingest/internal/resources/ranked_storage_account_set.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+
+	for _, tier := range accountsByTier {
+		rand.Shuffle(len(tier), func(i, j int) {
+			tier[i], tier[j] = tier[j], tier[i]
+		})
+	}
+
+	var result []RankedStorageAccount
```
AsafMah, line 98: You can pre-allocate result for a bit of efficiency - make([]RankedStorageAccount, 0, len(accountsByTier) * some constant (or calculate the length before))

kusto/ingest/internal/resources/ranked_storage_account_set_test.go
```diff
@@ -0,0 +1,228 @@
...
+	r := newDefaultRankedStorageAccountSet()
+
+	// Add 3 accounts
+	r.registerStorageAccount("test-account-1")
+	r.registerStorageAccount("test-account-2")
+	r.registerStorageAccount("test-account-3")
+
+	accounts := r.getRankedShuffledAccounts()
+
+	if len(accounts) != 3 {
```
AsafMah, line 20: Prefer using the testify library instead, there are plenty of shorthand function and it displays diffs better:

	assert.Len(t, accounts, 3)

kusto/ingest/internal/queued/queued.go
```diff
@@ -105,50 +105,84 @@ func New(db, table string, mgr *resources.Manager, http *http.Client, options ..
...
-	}
+	// Go over all the containers and try to upload the file to each one. If we succeed, we are done.
+	for attempts, containerUri := range containers {
+		if attempts >= StorageMaxRetryPolicy {
+			return errors.ES(errors.OpFileIngest, errors.KBlobstore, "max retry policy reached").SetNoRetry()
+		}
 
-	if err := i.Blob(ctx, blobURL, size, props); err != nil {
-		return err
+		client, containerName, err := i.upstreamContainer(containerUri)
```
AsafMah, line 138: Do you think abstracting the loops is beyond the scope or doesn't have enough benefit? I can can be convinced either way

kusto/ingest/internal/queued/queued.go
```diff
@@ -105,50 +105,84 @@ func New(db, table string, mgr *resources.Manager, http *http.Client, options ..
...
+		}
+
+		blobURL, size, err := i.localToBlob(ctx, from, client, containerName, &props)
+		if err == nil {
+			i.mgr.ReportStorageResourceResult(containerUri.Account(), true)
+			return i.Blob(ctx, blobURL, size, props)
+		}
+
+		// check if the error is retryable
+		if errors.Retry(err) {
```
AsafMah, line 151: is checking if the error is retryable only in one place intentional?

alonadam: Yes, localToBlob(...) can throw IO errors which are not related to storage, these errors are set to NoRetry.

kusto/ingest/internal/resources/ranked_storage_account_set.go
alonadam: I optimized by merging the code into 1 loop. Pre-allocate messes up the results (you end up with empty elements + the elements that were appended by append(...)).

kusto/ingest/internal/queued/queued.go
alonadam: for sure it is nice-to-have but I would like to finish this PR and also, I don't think it is that important.

kusto/ingest/internal/resources/resources.go
```diff
@@ -44,7 +44,7 @@ type URI struct {
 	sas                             url.Values
 }
 
-// Parse parses a string representing a Kutso resource URI.
+// parses a string representing a Kutso resource URI.
```
AsafMah, line 47: You need the name of the function in the start of the doc (one of those go things)

so it should be

// Parse parses a string ...

AsafMah (approved):
Great - just fix the one thing

PR #232, opened 2024-04-01
1.0.0 fixes

AsafMah:
### Changed
- [BREAKING] The Dynamic type now returns a []byte of json, it's up to the user to marshall it to the desired type. It can also be null.
- E2E tests now avoid creating tables when possible.
### Added
- Methods for getting a guid value (were missing).
### Fixed
- Fixed a panic when using the xByY methods on a null value.
- Fixed race in closing the client.


azkustodata/query/v2/iteartive_dataset.go
```diff
@@ -43,6 +43,7 @@ func NewIterativeDataset(ctx context.Context, r io.ReadCloser, capacity int) (qu
 	}
 
 	go func() {
+		defer d.reader.Close()
```
erankor, line 46: i think this reader leaks if prepareReadBuffer fails

azkustodata/value/value.go
```diff
@@ -59,7 +59,7 @@ func (p *pointerValue[T]) String() string {
 
 func (p *pointerValue[T]) GetValue() interface{} {
 	if p.value == nil {
-		return nil
+		return (*T)(nil)
```
erankor, line 62: can just remove the if, and always do return p.value

azkustodata/value/value.go
```diff
@@ -81,7 +81,7 @@ func parseError(expected interface{}, actual interface{}, err error) error {
 
 func (p *pointerValue[T]) Unmarshal(i interface{}) error {
 	if i == nil {
-		p.value = nil
+		p.value = (*T)(nil)
```
erankor, line 84: i don't think this is required, since value is typed as *T, can probably just leave as it was

azkustodata/query/v2/iteartive_dataset.go
AsafMah: Great catch!

azkustodata/query/v1/table.go
```diff
@@ -31,10 +31,12 @@ func NewTable(d query.BaseDataset, dt *RawTable, index *TableIndexRow) (query.Ta
 	columns := make([]query.Column, len(dt.Columns))
 
 	for i, c := range dt.Columns {
-		columns[i] = query.NewColumn(i, c.ColumnName, types.Column(c.ColumnType))
-		if !columns[i].Type().Valid() {
+		normal := types.NormalizeColumn(c.ColumnType)
+		if normal == "" {
 			return nil, errors.ES(op, errors.KClientArgs, "column[%d] if of type %q, which is not valid", i, c.ColumnType)
```
alonadam, line 36: wrong spelling - "is of type"

azkustodata/query/v2/iterative_table.go
```diff
@@ -73,10 +73,12 @@ func NewIterativeTable(dataset query.BaseDataset, th TableHeader) (query.Iterati
 
 func parseColumns(th TableHeader, columns []query.Column, op errors.Op) *errors.Error {
 	for i, c := range th.Columns() {
-		columns[i] = query.NewColumn(i, c.ColumnName, types.Column(c.ColumnType))
-		if !columns[i].Type().Valid() {
+		normal := types.NormalizeColumn(c.ColumnType)
+		if normal == "" {
 			return errors.ES(op, errors.KClientArgs, "column[%d] if of type %q, which is not valid", i, c.ColumnType)
```
alonadam, line 78: "is of type"

alonadam (approved):

PR #234, opened 2024-04-02
exponse the response headers

erankor:
- wrap the body reader + responseHeaders in a new query.Response
- add a new getter for the response headers on BaseDataset


AsafMah:
As we discussed in the mail, we don't want to implement this feature currently.
I'm closing for now, we can discuss it further.

PR #235, opened 2024-04-02
app and client tracing params

t-ronmoneta:
### Added
app and user for tracing fields in ingestion properties
Took those fields from KCSB and added them to ingestion Propeties in queued ingest client

### DIFF
--- a/azkustoingest/ingest.go
+++ b/azkustoingest/ingest.go
@@ -33,6 +33,8 @@ type Ingestion struct {
 
 	withoutEndpointCorrection    bool
 	customIngestConnectionString *azkustodata.ConnectionStringBuilder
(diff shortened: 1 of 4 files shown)


azkustoingest/ingest.go
```diff
@@ -45,6 +47,9 @@ func New(kcsb *azkustodata.ConnectionStringBuilder, options ...Option) (*Ingesti
 		kcsb = &newKcsb
 	}
 
+	i.applicationForTracing = kcsb.ApplicationForTracing
```
AsafMah, line 50: The default will be incorrect here - again, use ClientDetails

AsafMah (approved):

PR #248, opened 2024-05-30
1.0.0 - sync errors

AsafMah:
### Fixed
- Fixed various deadlocks and race conditions

### DIFF
--- a/azkustodata/query/v2/iterative_table.go
+++ b/azkustodata/query/v2/iterative_table.go
@@ -1,11 +1,13 @@
 package v2
 
 import (
+	"context"
 	"github.com/Azure/azure-kusto-go/azkustodata/errors"
 	"github.com/Azure/azure-kusto-go/azkustodata/query"
 	"github.com/Azure/azure-kusto-go/azkustodata/types"
 	"github.com/Azure/azure-kusto-go/azkustodata/value"
 	"sync"
+	"time"
 )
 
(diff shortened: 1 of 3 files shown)


azkustodata/query/v2/iterative_table.go
```diff
@@ -96,32 +100,71 @@ func parseRow(r []interface{}, t query.BaseTable, index int) (query.Row, *errors
...
+func (t *iterativeTable) reportError(err error) bool {
+	select {
+	case t.rows <- query.RowResultError(err):
+		return true
+	case <-t.ctx.Done():
+		return false
+	}
+}
+
+func (t *iterativeTable) tryReportError(err error) bool {
```
ohadbitt, line 127: Why do you need a ticker here - it seems to me theres no IO call - right?

azkustodata/query/v2/iterative_table.go
```diff
@@ -96,32 +100,71 @@ func parseRow(r []interface{}, t query.BaseTable, index int) (query.Row, *errors
...
+	ticker := time.NewTicker(100)
+	defer ticker.Stop()
+
+	select {
+	case t.rows <- query.RowResultError(err):
+		return true
+	case <-t.ctx.Done():
+		return false
+	case <-ticker.C:
```
ohadbitt, line 136: if ticker.C -> shouldnt we return error?

PR #255, opened 2024-07-24
feat(kcsb): support use of managed identity resource id

sgoings:
feat(kcsb): support use of managed identity resource id

deprecate WithUserManagedIdentity, with more specific replacements:
  - `WithUserAssignedIdentityClientId` - Receives the MSI client id
  - `WithUserAssignedIdentityResourceId` - Receives the MSI resource id

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `WithAppCertificatePath` - Receives the path to the certificate file.
   - `WithAppCertificateBytes` - Receives the certificate bytes in-memory.  
   Both methods accept an optional password for the certificate.
+- `WithUserManagedIdentity` has been deprecated in favor of more specific functions:
+  - `WithUserAssignedIdentityClientId` - Receives the MSI client id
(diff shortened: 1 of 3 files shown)


AsafMah:
Making it a heuristic is not great, it's probably better to have it as an explicit different parameter.

sgoings:
> Making it a heuristic is not great, it's probably better to have it as an explicit different parameter.

The way I'd see this working best is if `WithUserAssignedIdentity(...)` takes a struct `ManagedIdentityOpts` with fields `AppID` or `ResourceID` and then storing that struct as `ManagedServiceIdentity` in `ConnectionStringBuilder`, but not sure how willing you are to break API compat there.

Another way to handle it could be to have another function (such as `WithUserAssignedIdentityV2(opts ManagedIdentityOpts)`) and not change behavior of the existing `WithUserAssignedIdentity` function.

What direction would you prefer me take this?

AsafMah:
I think we should deprecate the old method, add two new methods - WithUserAssignedIdentityClientId and WithUserAssignedIdentityResourceId, and have old one redirect to the client id method.

sgoings:
@AsafMah ptal

AsafMah (approved):

PR #278, opened 2025-01-05
Refactor `TableClient` to use Azure Data Tables SDK

AsafMah:
- Replace deprecated `storage` package with `aztables` for table client operations.
- Update `NewTableClient` to initialize an `aztables.Client` using the new SDK.
- Modify `Read` and `Write` methods to utilize `aztables` API.
- Add support for context propagation in `Read` and `Write` methods.
- Serialize and deserialize entity data with JSON for compatibility with `aztables`.
- Remove outdated table service and entity handling logic.


azkustoingest/internal/status/status_table_client.go
```diff
@@ -1,67 +1,86 @@
...
+
+	service storage.TableServiceClient
+	table   *storage.Table
 }
 
 // NewTableClient Creates an azure table client.
-func NewTableClient(uri resources.URI) (*TableClient, error) {
-	c, err := storage.NewAccountSASClientFromEndpointToken(uri.URL().String(), uri.SAS().Encode())
+func NewTableClient(client policy.Transporter, uri resources.URI) (*TableClient, error) {
+	cred, err := aztables.NewClientWithNoCredential(uri.String(), &aztables.ClientOptions{
```
yogilad, line 31: This returns a client, so why name it `cred`?

azkustoingest/internal/status/status_table_client.go
```diff
@@ -1,67 +1,86 @@
...
-func (c *TableClient) Write(ingestionSourceID string, data map[string]interface{}) error {
-	var emptyID = uuid.Nil.String()
-	entity := c.table.GetEntityReference(ingestionSourceID, emptyID)
-	entity.Properties = data
-
-	options := &storage.EntityOptions{}
-	options.Timeout = defaultTimeoutMsec
+// Write reads a table record containing ingestion status.
+func (c *TableClient) Write(ctx context.Context, ingestionSourceID string, data map[string]interface{}) error {
+	dataCopy := make(map[string]interface{})
```
yogilad, line 66: Why does this require coping?

yogilad (approved):
Some comments, but looks ok

azkustoingest/internal/status/status_table_client.go
AsafMah: client is already a param name, I renamed it to tableClient now.

azkustoingest/internal/status/status_table_client.go
AsafMah: Technically it's surprising behavior if it modifies the data it received, but since we call it with a map that we throw away anyway there is no point of copying, so I removed the copy.

PR #285, opened 2025-02-12
Make `Timespan.Marshal()` follow Microsoft Standard TimeSpan "C" format

### LINKED ISSUE #284: `Timespan.Marshal()` does not follow Microsoft Standard TimeSpan "C" format
**Describe the bug**
`Timespan.Marshal()` yields wrong results for some values

**To Reproduce**
```go
NewTimespan(30 * time.Second).Marshal() // yields "00:00:3" instead of "00:00:30"
NewTimespan(1 * time.Millisecond).Marshal() // yields "00:00:00.001" instead of "00:00:00.0010000"
NewTimespan(1 * tick).Marshal() // yields "00:00:00.0001" instead of "00:00:00.0000001"
```

noamcohen97:
fixes #284

### DIFF
--- a/azkustodata/value/value_test.go
+++ b/azkustodata/value/value_test.go
@@ -3,7 +3,6 @@ package value
 import (
 	"encoding/json"
 	"math"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


AsafMah (approved):

azkustodata/value/value_test.go
```diff
@@ -2,10 +2,11 @@ package value
 
 import (
 	"encoding/json"
+	"fmt"
```
AsafMah, line 5: You don't have time, but have fmt twice

noamcohen97:
@AsafMah, Thanks for reviewing!

PR #291, opened 2025-03-11
Add sorting of query parameters by key in ToDeclarationString method.

### LINKED ISSUE #292: Parameters.ToDeclarationString is not deterministic
**Is your feature request related to a problem? Please describe.**
I'm using the query result cache (query_results_cache_max_age), and I noticed that my queries, despite having the same query text and parameters, are not equal. This discrepancy arises because the declaration string for the query parameters is not deterministic; iterating over a map results in a different order.

**Describe the solution you'd like**
Altering the function ToDeclarationString to produce a deterministic output when given the same parameters.

yonihorn:
Fixes #292


AsafMah:
Great contribution, thanks!

PR #306, opened 2025-04-27
Change Wait method to return a receive-only channel

noamcohen97:


AsafMah (approved):

AsafMah:
Thank you for the contribution!

PR #319, opened 2025-12-18
Fix `Ingestion.Reader` retry logic to reset buffer between attempts

### LINKED ISSUE #318: Ingestion.Reader retry logic doesn't reset buffer, causing ingestion failures or partial uploads
**Describe the bug**
The retry logic in `Ingestion.Reader` doesn't properly reset state between attempts, leading to ingestion failures or partial uploads.

1. Gzip compressor created outside retry loop: When compression is enabled, `gzip.Compress(reader)` is called once before the loop. After a failed upload consumes the stream, retries reuse the exhausted gzip stream.
  2. Reader not reset between retries: On failure, the code calls continue without seeking the reader back to the start. Retry attempts have no data to upload since the reader was already consumed.

noamcohen97:
Resolves #318

### DIFF
--- a/azkustoingest/internal/queued/queued_test.go
+++ b/azkustoingest/internal/queued/queued_test.go
@@ -5,17 +5,20 @@ import (
 	"compress/gzip"
 	"context"
 	"fmt"
-	"github.com/Azure/azure-kusto-go/azkustoingest/ingestoptions"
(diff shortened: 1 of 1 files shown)


AsafMah (approved):
Looks great!

PR #336, opened 2026-04-14
Fix queued compression metadata handling

mkeesey:
### Fixed

GenBlobName in non-compression cases would incorrectly override the blob extension it had just determined in an if-statement above. Compression discovery also prefers the Source.CompressionType configuration before falling back to discovery.

This allows use cases of pre-compressing data with more efficient gzip writers along with configuring azkustoingest.DontCompress() + azkustoingest.CompressionType(ingestoptions.GZIP)


asaharn:
Changes taken in #338

PR #338, opened 2026-04-21
[Fix]:  Queued compression metadata handling

asaharn:
Superseding https://github.com/Azure/azure-kusto-go/pull/336 by @mkeesey 

### Fixed
GenBlobName in non-compression cases would incorrectly override the blob extension it had just determined in an if-statement above. Compression discovery also prefers the Source.CompressionType configuration before falling back to discovery.

This allows use cases of pre-compressing data with more efficient gzip writers along with configuring azkustoingest.DontCompress() + azkustoingest.CompressionType(ingestoptions.GZIP)


mkeesey (approved):
Changes make sense to me - apologies for missing the streamed ingestion path as well.

tanmaya-panda1 (approved):

PR #349, opened 2026-08-21
Always close the underlying response body when decompression fails

### LINKED ISSUE #348: Response body is never closed when a compressed body read fails
When the response is compressed, `TranslateBody` wraps `resp.Body` in `originalCloser`, which closes the decompressor first and only closes the underlying body if that succeeded.

The order of operations on a failed read:

1. The body read fails mid-stream (e.g. the request context is canceled while results are streaming)
2. The decompressor's `Close` returns an error for the incomplete stream
3. `originalCloser.Close` returns early on that error, so `resp.Body.Close()` is never called

The problem is that every query that fails mid-download **leaks the HTTP response body unclosed**. On HTTP/1 this leaks the connection, and on HTTP/2, when the request context is not canceled, it can leak a stream slot on the shared connection.

noamcohen97:
Fixes #348

`originalCloser.Close` returned early when the decompressor's `Close` failed (which it does after any mid-stream read error), leaving the underlying `resp.Body` unclosed. This closes both unconditionally and joins the errors. Includes a regression test and unit tests for `originalCloser`.

### DIFF
--- a/azkustodata/internal/response/response.go
+++ b/azkustodata/internal/response/response.go
@@ -3,6 +3,7 @@ package response
 import (
 	"compress/flate"
 	"compress/gzip"
+	stderrors "errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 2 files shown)


tanmaya-panda1 (approved):

noamcohen97:
Thanks for reviewing and merging @tanmaya-panda1.
Is a new release with this fix planned anytime soon?

